PR #128, opened 2020-07-04
feat: Support english dialects

epage:
By default, we will accept all dialects but can be configured to only
allow one, correcting other dialects to it.

- [ ] Remove dialect corrections from dictionary
- [ ] Have dictionary correct to specific dialect


epage:
Might run into problems with https://github.com/rust-lang/rust/issues/74931

PR #224, opened 2021-04-18
perf(tokenize): Skip word splitting where possible

epage:


epage:
Just one sample of benchmarks that is representative of the whole
```
tokenize/ident/code     time:   [20.505 us 20.658 us 20.818 us]
                        change: [+470.99% +491.37% +516.56%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 9 outliers among 100 measurements (9.00%)
  5 (5.00%) high mild
  4 (4.00%) high severe
tokenize/words/code     time:   [1.4058 us 1.4102 us 1.4150 us]
                        change: [-1.4385% -0.7066% +0.1660%] (p = 0.09 > 0.05)
                        No change in performance detected.
```

Adding and checking for these groups is significantly slowing us down.  It appears a regex parser is also more costly than a fairly similar hand-rolled parser.

Comparing the ident vs word performance on master
```
tokenize/ident/code     time:   [3.5946 us 3.6107 us 3.6281 us]
tokenize/words/code     time:   [1.4160 us 1.4233 us 1.4321 us]
```

PR #267, opened 2021-05-31
Adding github action to use typos to check spelling

vsoch:
Hi @epage!

I really love this little tool, and wanted an easy way to add it to CI for some of my repos to check my spelling. I thought it might be useful to have a GitHub action, so this pull request is adding one here! The user will be able to add a step to a GitHub Workflow that looks like:

```yaml
    - name: Test extend identifiers
      uses: crate-ci/typos@<branch, commit, or relase>
      with: 
        files: ./file.txt
        force_pass: false
        extend_words: "word1,word2"
        extend_identifiers: "IdentifierToAdd,Secondidentifier"
```

And spell check some portion of their repository.  I've added a documentation page, and a GitHub Workflow to both demonstrate usage and test the action. Some points for discussion!

1. If you absolutely don't want it served alongside the code here, I'd be happy to put in its own repository. I just wanted to bring it here first because it's logical that the two will be synced.
2. The action builds directly from the Dockerfile added here, which installs the latest release of typos. This would need to be updated alongside releases. Another option would be to provide a docker image directly, which has the benefit of not needing to wait to build, but then you would need a separate release cycle for the container.  Building from source would probably take much longer, so I chose this middle of the road approach to build but use the release binary.
3. If you have preference for a different color or icon, or interface (what is printed) for the action, we can discuss that too!

Signed-off-by: vsoch <vsoch@users.noreply.github.com>

### DIFF
--- a/.github/workflows/test-action.yml
+++ b/.github/workflows/test-action.yml
@@ -1,0 +1,25 @@
+name: Test GitHub Action
+on: [pull_request]
+
+jobs:
+  run:
+    name: Spell Check with Typos
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout Actions Repository
+      uses: actions/checkout@v2
+
+    - name: Prepare file with mistakes.
+      run: echo "The quick brown foxx jumped over the slepy dog." > file.txt 
+    - name: Test force pass with mistakes
+      continue-on-error: true
+      uses: ./
+      with: 
+        files: ./file.txt
+
+    - name: Prepare file with no mistakes.
+      run: echo "The quick brown fox jumped over the sleepy dog." > file.txt 
+    - name: Test force pass with no mistakes
+      uses: ./
+      with: 
+        files: ./file.txt
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,8 @@
+FROM ubuntu:20.04
+ARG VERSION=1.0.3
+ENV VERSION=${VERSION}
+RUN apt-get update && apt-get install -y wget
+RUN wget https://github.com/crate-ci/typos/releases/download/v${VERSION}/typos-v${VERSION}-x86_64-unknown-linux-gnu.tar.gz && \
+    tar -xzvf typos-v${VERSION}-x86_64-unknown-linux-gnu.tar.gz && \
+    mv typos /usr/local/bin
+ENTRYPOINT ["/usr/local/bin/typos"]
--- a/README.md
+++ b/README.md
@@ -20,6 +20,10 @@ Dual-licensed under [MIT](LICENSE-MIT) or [Apache 2.0](LICENSE-APACHE)
 
 - [Installation](#install)
 - [Getting Started](#getting-started)
+  - [False Positives](#false-positives)
+  - [Integrations](#integrations)
+    - [GitHub Action](docs/github-action.md)
+  - [Debugging](#debugging)
 - [Reference](docs/reference.md)
 - [Comparison with other spell checkers](docs/comparison.md)
 - [Benchmarks](benchsuite/runs)
@@ -36,7 +40,6 @@ Or use rust to install:
 ```bash
 cargo install typos-cli
 ```
-
 ## Getting Started
 
 Most commonly, you'll either want to see what typos are available with
@@ -82,7 +85,7 @@ extend-exclude = ["localized/*.po"]
 
 ### Integrations
 
-`typos` provides several building blocks for custom integrations
+`typos` provides several building blocks for custom native integrations
 - `-` reads from `stdin`, `--write-changes` will be written to `stdout`
 - `--diff` to provide a diff
 - `--format json` to get jsonlines with exit code 0 on no errors, code 2 on typos, anything else is an error.
@@ -96,6 +99,9 @@ typos dir/file --diff
 # Fully programmatic control
 typos dir/file --format json
 ```
+
+In addition, see the documentation on [GitHub Actions](docs/github-actions.md),
+another integration.
 
 ### Debugging
 
--- a/docs/github-action.md
+++ b/docs/github-action.md
@@ -1,0 +1,48 @@
+# GitHub Action
+
+If you want an easy way to test your repository spelling (or a subset of files)
+you can use the Typos Action! It is served from this repository, and can
+easily be used as follows:
+
+```yaml
+name: Test GitHub Action
+on: [pull_request]
+
+jobs:
+  run:
+    name: Spell Check with Typos
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout Actions Repository
+      uses: actions/checkout@v2
+
+    - name: Check spelling of file.txt
+      uses: crate-ci/typos@main
+      with: 
+        files: ./file.txt
+
+    - name: Use custom config file
+      uses: crate-ci/typos@main
+      with: 
+        files: ./file.txt
+        config: ./myconfig.toml
+
+    - name: Ignore implicit configuration file
+      uses: crate-ci/typos@main
+      with: 
+        files: ./file.txt
+        isolated: true
+```
+
+**Important** for any of the examples above, make sure that you choose
+a release or commit as a version, and not a branch (which is a moving target).
+Also make sure when referencing relative file paths to use `./` (e.g., `./file.txt` instead of
+`file.txt`.
+
+## Variables
+
+| Name | Description | Required | Default |
+|------|-------------|----------|---------|
+| files| Files or patterns to check | false | If not defined, entire repository is checked |
+| isolated | Ignore implicit configuration files | false | false|
+| config | Use a custom config file (must exist) | false | not set |
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ pre-release-replacements = [
   {file="CHANGELOG.md", search="ReleaseDate", replace="{{date}}", min=1},
   {file="CHANGELOG.md", search="<!-- next-header -->", replace="<!-- next-header -->\n## [Unreleased] - ReleaseDate\n", exactly=1},
   {file="CHANGELOG.md", search="<!-- next-url -->", replace="<!-- next-url -->\n[Unreleased]: https://github.com/assert-rs/predicates-rs/compare/{{tag_name}}...HEAD", exactly=1},
+  {file="docker/Dockerfile", search="ARG VERSION=.*", replace="ARG VERSION={{version}}", min=1},
 ]
 
 [[bin]]
--- a/action.yml
+++ b/action.yml
@@ -1,0 +1,33 @@
+name: "typos-action"
+author: "Vanessa Sochat"
+description: "Run typos to check spelling in GitHub actions"
+
+inputs:
+  files:
+    description: "Files or patterns to check"
+    required: false
+
+  extend_identifiers:
+    description: "Comma separated list of extend identifiers, like someone's name"
+    required: false
+
+  extend_words:
+    description: "Comma separated list of extend words."
+    required: false
+
+  isolated:
+    description: "Ignore implicit configuration files"
+    required: false
+    default: false
+
+  config:
+    description: "Use a custom config file."
+    required: false
+
+runs:
+  using: "docker"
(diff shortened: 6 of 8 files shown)


Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM ubuntu:20.04
+ARG VERSION=1.0.3
```
epage, line 2: We could [ask github what the latest release is](https://github.com/crate-ci/gh-install/blob/master/v1/install.sh#L119)

.github/workflows/test-action.yml
```diff
@@ -0,0 +1,43 @@
...
+    name: Spell Check with Typos
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout Actions Repository
+      uses: actions/checkout@v2
+
+    - name: Test force pass with mistakes
+      uses: ./
+      with: 
+        files: ./src/file.rs
```
epage, line 15: We should generate this file so this doesn't stop testing things for us if I refactor `file.rs`

Maybe a `test/typo.txt` file.  I could then have `src/file` and other places source that as well and we could then have it generally ignored, allowing running `typos` on `typos`.

README.md
```diff
@@ -20,6 +20,7 @@ Dual-licensed under [MIT](LICENSE-MIT) or [Apache 2.0](LICENSE-APACHE)
 
 - [Installation](#install)
 - [Getting Started](#getting-started)
+- [GitHub Action](docs/github-action.md)
```
epage, line 23: Should we instead reference this down in the "Integrations" section?

Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM ubuntu:20.04
```
epage, line 1: Should we use a lighter weight `FROM`?

epage (commented):
Thanks for doing this! This is definitely

This is my first foray into providing an action and would appreciate your thoughts on my comments!

Maybe at a later stage we can upgrade this to posting comments inline to the diff view.

Dockerfile
epage: Or better, add a [pre-release replacement](https://github.com/crate-ci/typos/blob/master/Cargo.toml#L32) to auto-update this on release.

docs/github-action.md
```diff
@@ -0,0 +1,65 @@
...
+`file.txt`. You can also define identifiers and words to skip.
+
+## Variables
+
+| Name | Description | Required | Default |
+|------|-------------|----------|---------|
+| files| Files or patterns to check | false | If not defined, entire repository is checked |
+| extend_identifiers | Comma separated list of extend identifiers, like someone's name | false | not set |
+| extend_words | Comma separated list of extend words. | false | not set |
+| force_pass | Force pass of check, regardless of check | false | false |
```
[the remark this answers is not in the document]
vsoch: That's a good point - the only detail really is that you get different messages from the tool depending on this variable. E.g.,

```bash
# mistakes found, force_pass = false
Spelling mistakes found! 😱️

# mistakes found, force_pass = true
Spelling mistakes found, but passing anyway! 😚️

# no mistakes!
No spelling mistakes found! 🎉️
```
If you feel strongly about it that it's redundant, I can remove it. But I suspect most people aren't super familiar with GitHub actions syntax and would more quickly / easily use force_pass.

docs/github-action.md
```diff
@@ -0,0 +1,65 @@
...
+referencing relative file paths to use `./` (e.g., `./file.txt` instead of
+`file.txt`. You can also define identifiers and words to skip.
+
+## Variables
+
+| Name | Description | Required | Default |
+|------|-------------|----------|---------|
+| files| Files or patterns to check | false | If not defined, entire repository is checked |
+| extend_identifiers | Comma separated list of extend identifiers, like someone's name | false | not set |
+| extend_words | Comma separated list of extend words. | false | not set |
```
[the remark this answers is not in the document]
vsoch: The idea here is that these variables are written to a config file, but just in the action, so the user can add the logic of the check completely to the action without needing to add an extra file to the repository. That said, if the config file already exists it is honored (and these are ignored).

I'll add isolated and config!

Dockerfile
```diff
@@ -0,0 +1,10 @@
...
+ARG VERSION=1.0.3
+ENV VERSION=${VERSION}
+RUN apt-get update && apt-get install -y git wget
+RUN wget https://github.com/crate-ci/typos/releases/download/v${VERSION}/typos-v${VERSION}-x86_64-unknown-linux-gnu.tar.gz && \
+    tar -xzvf typos-v${VERSION}-x86_64-unknown-linux-gnu.tar.gz && \
+    mv typos /usr/local/bin
+COPY entrypoint.sh /entrypoint.sh
+WORKDIR /github/workspace
+ENTRYPOINT ["/bin/bash", "/entrypoint.sh"]
```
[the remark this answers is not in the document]
vsoch: That's a good point! I can a base typos container to the root, and for now a (mostly duplicate) version for GitHub actions in a docker subfolder. Then if you want to provide a container base with typos, it would b possible to set up an automated build and deploy for the container in the root, and have the actions use it as a base.

Dockerfile
vsoch: I think we actually would want the container to be pinned with the current release for the repo - the reason is that a user can request an action for a specific release version, e.g.:

```yaml
uses: crate-ci/typos@1.0.3
```
If we have the container always use the latest, the release versions for the action won't have any meaning.

vsoch: Hmm the pre-release replacement is a cool idea... but it would mean only releases will work for the GitHub action (and not branches). Using branches isn't best practice but it's fairly common!

.github/workflows/test-action.yml
vsoch: What about generating the file on the fly like the other tests? It makes it easy to have tiny differences. And if you decide to use the text in other places it could be written to file!

README.md
vsoch: Sure thing!

vsoch: Actually looking at the section I think I disagree - this is talking more about command line usage of typos. GitHub Actions I think is going to be the only interest for some folks, and if we just remove it there and hope they read down further, they won't see it. So I think we should keep it at the top level as a separate thing so that someone coming to the repo _looking_ for it explicitly sees it quickly.

Dockerfile
vsoch: Sure, what base would you like to use?

Dockerfile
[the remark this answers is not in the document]
vsoch: We can also delete the Dockerfile from the root now, and keep the actions content hidden in the docker/ subfolder. If you decide to add a container for typos with automated builds you could add it back. Let me know your preference!

docs/github-action.md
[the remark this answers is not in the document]
epage: Unless there is a standing convention for this having a `force_pass`, I'd rather rely on the built-in method.

docs/github-action.md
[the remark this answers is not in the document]
epage: If people have need for specifying this config, they should be providing the config file.  This becomes a hinderance to the developer or contributors because how they run `typos` locally won't work with how it is run in CI.

Now, I could see us adding support for pulling the config from project files, e.g. `Cargo.toml` or `pyproject.toml`s metadata / tool tables.

Dockerfile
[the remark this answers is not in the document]
epage: I can go either way but if we are getting rid of the `extend` config, I suspect we can get rid of the entry point.

Dockerfile
epage: Im not sure why.  Pre-release replacements just automatically update the `Dockerfile` in master when making a new release.  Branches can point to whatever version the branch author feels is appropriate.

.github/workflows/test-action.yml
epage: I think thats fine.  I was just adding another option to try to isolate these test cases so its easier to run typos on itself.

README.md
epage: Its talking about that, because that is all there is for now.  I'd put native integrations, like Github Actions, before instuctions for custom integrations (seeing as this technically would be building on the stuff documented in the custom integrations)

Dockerfile
epage: Not an expert but I hear Alpine is always good ... except for Python projects.

Otherwise, we can keep ubuntu.  If the runner is already using that image, then its already pulled and ready to go.

docs/github-action.md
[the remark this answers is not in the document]
vsoch: okay, will remove!

docs/github-action.md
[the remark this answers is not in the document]
vsoch: okay, we can just remove those extra arguments then. I'm not sure I'll use the action but I'm sure others will!

Dockerfile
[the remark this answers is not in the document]
vsoch: I don't follow here - I'll push changes in a little bit and we can discuss what you'd like.

Dockerfile
vsoch: Could you show me how you would implement this? From what I can see, the pre-release replacement is changing a string to a version. This means that (unless the string already is a version, which defeats the purpose in the first place) it would mean that the Dockerfile in master won't build. I'll push changes soon and you can make a suggestion on the file to show me what you would like (I don't think I follow completely).

.github/workflows/test-action.yml
vsoch: Great!

README.md
vsoch: okay, I'll try to extend the table of contents to  have GitHub actions under integrations.

Dockerfile
vsoch: I gave alpine a shot, but typos didn't work inside. Here is the Dockerfile if you want to give it a try!

```
FROM alpine:3.13
ARG VERSION=1.0.3
ENV VERSION=${VERSION}
RUN apk update && apk add wget
RUN wget https://github.com/crate-ci/typos/releases/download/v${VERSION}/typos-v${VERSION}-x86_64-unknown-linux-gnu.tar.gz && \
    tar -xzvf typos-v${VERSION}-x86_64-unknown-linux-gnu.tar.gz && \
    mv typos /usr/local/bin
ENTRYPOINT ["/usr/local/bin/typos"]
```

epage: Its ok. I need to go modify the Linux builds to include musl statically firt.

Dockerfile
epage: In `Cargo.toml`. add the line
```toml
  {file="docker/Dockerfile", search="ARG VERSION=.*", replace="ARG VERSION={{version}}", min=1},
```
So when I release 1.0.4, `cargo release` will automatically update the Dockerfile` to 1.0.4.

Dockerfile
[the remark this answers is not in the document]
epage: Its moved, its ok.  We can always iterate on this later ayways

Dockerfile
vsoch: ohh I see! I totally misread the command to think that the `{{ version }}` needed to be in the Dockerfile directly (like a jinja2 template) but I see that it's referencing the actual version that is there. Yes this is a great idea, will update the PR to have this.

vsoch: All set.

epage:
Thanks!

vsoch:
Woot! What happened with the revert? 

I'll watch the repo for releases, and grab the next release to make sure the action is working ok after that. Then I'll try to share it a little more broadly.

epage:
> What happened with the revert?

I had misclicked and then misclicked again :)

> I'll watch the repo for releases, and grab the next release to make sure the action is working ok after that. Then I'll try to share it a little more broadly.

Working on a release but I seem to be having problems building atm.  I think PHF is killing my machine, so trying to reduce the load for building the crate.

vsoch:
Works great! https://github.com/singularityhub/sregistry/pull/367. I added to my blog, and also put out a tweet (I couldn't find your Twitter username otherwise would have pinged you). I'll likely add to other projects as I realize how bad the spelling is :p Thanks for this great tool!

PR #289, opened 2021-06-28
feat(pre-commit): add binary based install

### LINKED ISSUE #285: [Pre-commit] Feature Request (Remove Crate-Dependency)
I really like this project and would like to include it as a pre-commit hook, but asking all my git repo to install Rust for the pre-commit hook is a bit much. As such, would it be possible to wrap the binary in a setup.py like is done by CircleCI-Py and such https://github.com/AleksaC/hadolint-py/blob/main/setup.py . That way it would not need any additional pre-reqs besides the ones for pre-commit.

scop:
For no additional build dependencies (besides Python >= 3.6), and faster
install.

The setup.py here is a copy of the one at
https://github.com/shellcheck-py/shellcheck-py with changes kept at the
bare minimum for future diffability.

We don't support checksumming at least yet, because there's no access to
them from cargo-release at the moment.

Closes #285

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,166 @@
+#!/usr/bin/env python3
+import hashlib
+import http
+import io
+import os.path
+import stat
+import sys
+import tarfile
+import urllib.request
+import zipfile
+from distutils.command.build import build as orig_build
+from distutils.core import Command
+from typing import Tuple
+
+from setuptools import setup
+from setuptools.command.install import install as orig_install
+
+TYPOS_VERSION = '1.0.9'
+POSTFIX_SHA256 = {
+    'linux': (
+        'x86_64-unknown-linux-gnu.tar.gz',
+        '',  # TODO: sha256 hexhash when we can generate it with release
+    ),
+    'darwin': (
+        'x86_64-apple-darwin.tar.gz',
+        '',  # TODO: sha256 hexhash when we can generate it with release
+    ),
+    'win32': (
+        'x86_64-pc-windows-msvc.zip',
+        '',  # TODO: sha256 hexhash when we can generate it with release
+    ),
+}
+PY_VERSION = '1'
+
+
+def get_download_url() -> Tuple[str, str]:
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -0,0 +1,166 @@
...
+import urllib.request
+import zipfile
+from distutils.command.build import build as orig_build
+from distutils.core import Command
+from typing import Tuple
+
+from setuptools import setup
+from setuptools.command.install import install as orig_install
+
+TYPOS_VERSION = '1.0.9'
```
epage, line 18: Please update `Cargo.toml`s `[package.metadata.release]` table to ensure we update this automatically on each release

setup.py
```diff
@@ -0,0 +1,166 @@
...
+POSTFIX_SHA256 = {
+    'linux': (
+        'x86_64-unknown-linux-gnu.tar.gz',
+        '',  # TODO: sha256 hexhash when we can generate it with release
+    ),
+    'darwin': (
+        'x86_64-apple-darwin.tar.gz',
+        '',  # TODO: sha256 hexhash when we can generate it with release
+    ),
+    'win32': (
```
epage, line 28: ... `sys.platform()` uses `win32` for both 32-bit and 64-bit :/

(this confused me at first and looked it up)

epage (commented):
Thanks for doing this!

epage:
And ignore the CI failures.  One is a CI bug (shouldn't be failing build, just the job) and the other seemed to have been broken by a new Rust release.

setup.py
scop: Gah, how did I forget about that, when I remembered and documented the limitation on sha256sums :thinking: :stuck_out_tongue: 

Hopefully fixed by ef76d20c6a8060fb1036d4b0db0d0c29d441ff82

epage:
I should add a `cargo release --dry-run` to CI to help catch problems with `release.toml`.  *shrug*

PR #330, opened 2021-08-20
Ignore hexadecimal "hashes" of length 32 or greater.

### LINKED ISSUE #326: Find some way to tolerate hexadecimal better
I have a program that uses a fair amount of hex strings in order to represent binary sequences in its tests.  But this doesn't sit so well with `typos`, which gives lots of false positives like:

```
error: `ba` should be `by`, `be`
  --> ./tor-llcrypto/src/pk.rs:77:21
   |
77 |              0816ed13ba3303ac5deb911548908025"
```

It seems that this is a byproduct of treading any sequence of non-numeric wordlike characters as a possible word, even if they occur adjacent to numbers, not specified by spaces.

nmathewson:
By experimentation (see ticket), it seems that same-case hexadecimal
strings of 32 characters or longer are almost never intended to hold
text.  By treating such strings as ignored, we can resist a larger
category of false positives.

Closes #326.

### DIFF
--- a/crates/typos/src/tokens.rs
+++ b/crates/typos/src/tokens.rs
@@ -264,13 +264,20 @@ mod parser {
         <T as nom::InputIter>::Item: AsChar + Copy,
     {
         // Size considerations:
-        // - sha-1 is git's original hash
-        // - sha-256 is git's new hash
-        // - Git hashes can be abbreviated but we need a good abbreviation that won't be mistaken
-        //   for a variable name
-        const SHA_1_MAX: usize = 40;
-        const SHA_256_MAX: usize = 64;
-        take_while_m_n(SHA_1_MAX, SHA_256_MAX, is_lower_hex_digit)(input)
+        //   - 40 characters holds for a SHA-1 hash from older Git versions.
+        //   - 64 characters holds for a SHA-256 hash from newer Git versions.
+        //   - Git allows abbreviated hashes, but we need a good abbreviation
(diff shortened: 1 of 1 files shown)


crates/typos/src/tokens.rs
```diff
@@ -264,13 +264,16 @@ mod parser {
         <T as nom::InputIter>::Item: AsChar + Copy,
     {
         // Size considerations:
-        // - sha-1 is git's original hash
-        // - sha-256 is git's new hash
-        // - Git hashes can be abbreviated but we need a good abbreviation that won't be mistaken
-        //   for a variable name
-        const SHA_1_MAX: usize = 40;
-        const SHA_256_MAX: usize = 64;
```
epage, line 272: Please carry over the comments as they are also important context to carry forward,. though without the `const`s, you'll need to add to them the sizes for each hash

epage: Thanks!  Go ahead and fixup and re-push and we'll get this merged

(I wish Github had native support for `fixup`)

epage:
Looks like rustfmt is complaining.  Don't worry too much about the lint committer, that is more advisory

nmathewson:
Ran "cargo fmt"; squashed; re-pushed.

nmathewson:
Okay; looks green except for "lint commits"

epage:
Released

PR #374, opened 2021-11-12
Fix multiple escape sequences

Flakebi:
If escape sequences follow straight after each other, there is no
delimiter in-between.
In such a case, parsing previously stopped and did not find any
typos further in the file.

Also change the c-escape parser to take either \\+\<single char> or
\\+\<char sequence delimited by space or similar>.

Also update trycmd, so failures actually show up in CI.

### DIFF
--- a/crates/typos/src/tokens.rs
+++ b/crates/typos/src/tokens.rs
@@ -190,8 +190,8 @@ mod parser {
             terminated(base64_literal, sep1),
             terminated(email_literal, sep1),
             terminated(url_literal, sep1),
-            terminated(c_escape, sep1),
-            terminated(printf, sep1),
+            c_escape,
+            printf,
             sep1,
         )))(input)
     }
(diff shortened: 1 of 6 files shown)


tests/cmd/stdin-failure2.toml
```diff
@@ -0,0 +1,3 @@
+bin.name = "typos"
+args = "-"
+status.code = 2
```
epage, line 3: Should we make this case `--write-changes` so we can validate the `2` is for the reason we think?

Flakebi: Good idea

epage (commented):
Sorry, I thought these were published

crates/typos/src/tokens.rs
```diff
@@ -410,7 +410,12 @@ mod parser {
         <T as nom::InputTakeAtPosition>::Item: AsChar + Copy,
         <T as nom::InputIter>::Item: AsChar + Copy,
     {
-        preceded(take_while1(is_escape), take_while(is_xid_continue))(input)
+        // Match either a multi char escape sequence like \x01 until a separator
+        // or a single char escape like \<newline> or \<space>.
+        preceded(
+            take_while1(is_escape),
+            alt((take_while1(is_xid_continue), take(1usize))),
+        )(input)
```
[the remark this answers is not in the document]
Flakebi: That makes a lot of sense, I added that as a comment.

epage (approved):
Thanks!

epage:
1.3.0 is now released!

Flakebi:
Thank you!

PR #389, opened 2021-12-18
feat(pre-commit): Make work on cygwin

scop:
Borrowed from https://github.com/shellcheck-py/shellcheck-py/pull/42, thanks to @DragonCrafted87.

I have no idea if this actually works nor do I have an easy way to check, just happened to notice and thought it might.


epage:
Thanks!

Issue #427, opened 2022-02-05
Not available dockerhub/external repo for dockerfile


georgettica:
I found https://hub.docker.com/r/typoci/spellcheck-action

But it seems very old

If you vae something the README.md should reflex this to allow this very tool to be used outside of github-actions

epage:
That is a different project.

I have not published to docker hub because I've just no taken the time to
- Create a more general docker image (the current one is focused purely on the action)
- Looked into the process for uploading an image from CI (since I don't want to do it manually).

froque:
> Looked into the process for uploading an image from CI (since I don't want to do it manually).

I just learned that Docker can build the image themselves: https://docs.docker.com/docker-hub/builds/

weijiany:
Hi @epage  I noticed that there's already a [Dockerfile](https://github.com/crate-ci/typos/blob/master/Dockerfile) in the repository, but it seems that the image isn't being pushed to any external container registry, like Docker Hub or GHCR. Are there any plans to do this?

Having the image available in a public registry would be very helpful, especially for integrating this tool into other CI tools not only Github Action. It would streamline the process by allowing us to pull the image directly, rather than needing to build it ourselves.

epage:
@weijiany see https://github.com/crate-ci/typos/issues/427#issuecomment-1031618262 with #431 being the only update since then.

If there is a quick resource on getting this up and going where I do not need to intervene for each release, I will add it.

PR #431, opened 2022-02-11
New dockerfile for #427.

halkeye:
Builds all the binaries on demand, then creates a new image with just the binary.
Easy to do tagged releases or multi arch builds, etc.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
-FROM ubuntu:20.04
-ARG VERSION=1.4.0
-ENV VERSION=${VERSION}
-RUN apt-get update && apt-get install -y wget
-RUN wget https://github.com/crate-ci/typos/releases/download/v${VERSION}/typos-v${VERSION}-x86_64-unknown-linux-musl.tar.gz && \
-    tar -xzvf typos-v${VERSION}-x86_64-unknown-linux-musl.tar.gz && \
-    mv typos /usr/local/bin
-ENTRYPOINT ["/usr/local/bin/typos"]
+FROM rust:1.58.1 as builder
+WORKDIR /usr/src/typos
+COPY . .
+RUN cargo install --path .
+
+FROM debian:buster-slim
+COPY --from=builder /usr/local/cargo/bin/typos /usr/local/bin/typos
+CMD ["typos"]
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+/target
+Dockerfile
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -1,8 +1,8 @@
...
-    mv typos /usr/local/bin
-ENTRYPOINT ["/usr/local/bin/typos"]
+FROM rust:1.58.1 as builder
+WORKDIR /usr/src/typos
+COPY . .
+RUN cargo install --path .
+
+FROM debian:buster-slim
+COPY --from=builder /usr/local/cargo/bin/typos /usr/local/bin/typos
+CMD ["typos"]
```
halkeye, line 8: having it as command instead of entry point makes it more overriable, but less obvious

`docker run crate-ci/typos` would run typos
`docker run crate-ci/typos bash` would run bash
`docker run crate-ci/typos typos --with --args` would run typos

i can make a smarter entrypoint that handles bash vs not bash though.

https://devopscube.com/run-scripts-docker-arguments/

halkeye:
Does cargo install build the right one? I'll have to run again when I'm at a computer to confirm its the release version. Otherwise I can probably build release and copy from target instead

Dockerfile
```diff
@@ -1,8 +1,8 @@
...
-ENV VERSION=${VERSION}
-RUN apt-get update && apt-get install -y wget
-RUN wget https://github.com/crate-ci/typos/releases/download/v${VERSION}/typos-v${VERSION}-x86_64-unknown-linux-musl.tar.gz && \
-    tar -xzvf typos-v${VERSION}-x86_64-unknown-linux-musl.tar.gz && \
-    mv typos /usr/local/bin
-ENTRYPOINT ["/usr/local/bin/typos"]
+FROM rust:1.58.1 as builder
+WORKDIR /usr/src/typos
+COPY . .
+RUN cargo install --path .
```
epage, line 4: Since this is building from source, will actions pull in the right version?  Actions tend to be tied to a specific version so that should pull in a specific version of this image, so we should be fine.  If someone specifies `master` though, they'll get an unreleased version.  That is probably acceptable.

Dockerfile
```diff
@@ -1,8 +1,8 @@
-FROM ubuntu:20.04
-ARG VERSION=1.4.0
```
epage, line 2: Could you remove the line in `Cargo.toml` that updates this?

Dockerfile
```diff
@@ -1,8 +1,8 @@
...
-    tar -xzvf typos-v${VERSION}-x86_64-unknown-linux-musl.tar.gz && \
-    mv typos /usr/local/bin
-ENTRYPOINT ["/usr/local/bin/typos"]
+FROM rust:1.58.1 as builder
+WORKDIR /usr/src/typos
+COPY . .
+RUN cargo install --path .
+
+FROM debian:buster-slim
+COPY --from=builder /usr/local/cargo/bin/typos /usr/local/bin/typos
```
epage, line 7: So this is updating our general docker image but the `docker/Dockerfile` is the one used by actions.  We use a bash script to get extra logic to integrate with the action.

I'm asuming we should keep that separate and have it be a layer on top of this image.

halkeye: okay good, I was trying to read up on it. 

I think that should stay until the core image is published and tagged, then yea, either merge the two and use image in the action.yml, or use the tagged version as FROM

Dockerfile
halkeye: updated

halkeye:
could use https://github.com/docker/build-push-action/blob/master/docs/advanced/multi-platform.md to build and publish to ghcr.io

epage:
Thanks!

Issue #458, opened 2022-04-06
ignore go.sum by default


georgettica:
in the same way https://github.com/sharkdp/fd has a --no-ignore or --hidden flags, we should automatically not search `go.sum` for typos (as it's generated I think)

epage:
This is effectively a lock file, yes?

If we add it to our ["lock" file type](https://github.com/crate-ci/typos/blob/master/src/default_types.rs#L149), we'll then [stop checking the file](https://github.com/crate-ci/typos/blob/master/src/config.rs#L178)

epage:
Mind taking a look at https://github.com/crate-ci/typos/pull/459 to see if the description and actions taken line up with a go users expectations?

georgettica:
This looks great, thank you!

georgettica:
Testing soon

georgettica:
it works!
glad this was merged :)

georgettica:
@epage found this works great but I now have an issue in go.mod
this shows the imports that can be done, and sometimes we use a hash to describe a version..

this is less obvious to fix as we need to ignore the version from scanning
```
module example.com

go 1.18

require (
        foo.com/pkg somebadhash
)
```

we need to ignore the `somebadhash` in go.mod

epage:
Unless we do something special go.mod, this falls under https://github.com/crate-ci/typos/issues/415 which we've not had a good solution for yet.

PR #459, opened 2022-04-06
fix(config): Set file types for go.mod/go.sum 

epage:
`go.mod` seems to be a specification file which we tend to lump in with
the language itself since a weirdly spell dependency will likely show up
in code.

`go.sum` seems to be like a lock file which we quarantine into its own
file type.

Fixes https://github.com/crate-ci/typos/issues/458


georgettica:
Looks great!
I was excluding it until now 😆

PR #460, opened 2022-04-10
docs(reference): Formatting and copy-pasto fixes

scop:
Escape left angle bracket, fix copy-pastos in default.extend-identifiers and default.extend-words descriptions.


epage (approved):
Thanks!

PR #464, opened 2022-04-25
Fix tokenizer for uppercase UUID

foriequal0:
Microsoft toolchains usually emit UUID/GUID in UPPERCASE


epage (approved):
> Microsoft toolchains usually emit UUID/GUID in UPPERCASE

Of course they do (the worst was when the GUID generator created mixed-case GUIDs).

Implemented the current form off of wikipedia which said

> [RFC 4122 Section 3](https://tools.ietf.org/html/rfc4122#section-3) requires that the characters be generated in lower case, while being case-insensitive on input.

but willing to expand this out as the use case is large enough

epage:
Released in v1.7.1

PR #465, opened 2022-04-25
Fix misalignment in report highlight and indent when there are double width characters

foriequal0:
Previous method misaligns highlights when there are double width asian characters
```
39 |  한글 eglish
   |    ^^^^^^
```
This commit fixes the highlight to have correct alignment.
```
39 |  한글 eglish
   |      ^^^^^^
```
`unicode-rs` crate is used by the Rust compiler [1].
[1]: https://github.com/rust-lang/rust/blob/34a6c9f26e2ce32cad0d71f5e342365b09f4d12c/compiler/rustc_errors/src/emitter.rs#L861

I've intentionally left `column` variable as is since many IDE counts by grapheme clusters.

### DIFF
--- a/src/bin/typos-cli/report.rs
+++ b/src/bin/typos-cli/report.rs
@@ -3,6 +3,8 @@
 use std::io::{self, Write};
 use std::sync::atomic;
 
+use unicode_width::UnicodeWidthStr;
+
 use typos_cli::report::{Context, Message, Report, Typo};
 
 #[derive(Copy, Clone, Debug)]
@@ -218,8 +220,11 @@ fn print_long_correction(msg: &Typo, palette: Palette) -> Result<(), std::io::Er
         let line_num = context.line_num.to_string();
         let line_indent: String = itertools::repeat_n(" ", line_num.len()).collect();
 
-        let hl_indent: String = itertools::repeat_n(" ", column).collect();
-        let hl: String = itertools::repeat_n("^", msg.typo.len()).collect();
+        let visible_column = UnicodeWidthStr::width(start.as_ref());
+        let visible_len = UnicodeWidthStr::width(msg.typo);
+
+        let hl_indent: String = itertools::repeat_n(" ", visible_column).collect();
+        let hl: String = itertools::repeat_n("^", visible_len).collect();
 
         writeln!(handle, "{} |", line_indent)?;
         writeln!(
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1576,6 +1576,7 @@ dependencies = [
  "typos-vars",
  "unicase",
  "unicode-segmentation",
+ "unicode-width",
  "varcon-core",
  "yansi",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -91,6 +91,7 @@ encoding = "0.2"
 kstring = { version = "2.0.0", features = ["serde"] }
 typed-arena = "2.0.1"
 maplit = "1.0"
+unicode-width = "0.1.9"
 
 [dev-dependencies]
 assert_fs = "1.0"
--- a/tests/cmd/stdin-failure-multiwidth.stdin
+++ b/tests/cmd/stdin-failure-multiwidth.stdin
@@ -1,0 +1,5 @@
+한 Apropriate world
+한 Apropriate world
+😵‍💫 Apropriate world
+👁️‍🗨️ Apropriate world
+	Apropriate world
--- a/tests/cmd/stdin-failure-multiwidth.stdin
+++ b/tests/cmd/stdin-failure-multiwidth.stdin
@@ -1,5 +1,34 @@
+---
+Korean character (NFC)
+Grapheme clusters: 1, codepoints: 1, UnicodeWidthStr::width() == 2
+
 한 Apropriate world
+   ^^^^^^^^^^ highlight here
+
+---
+Korean character (NFD).
+Grapheme clusters: 1, codepoints: 3, UnicodeWidthStr::width() == 2
+
 한 Apropriate world
-😵‍💫 Apropriate world
+   ^^^^^^^^^^ highlight here
+
+---
+Eye in Speech Bubble Emoji (U+1F441 U+FE0F U+200D U+1F5E8 U+FE0F, Recommended Emoji ZWJ Sequences, v2.0)
+Grapheme clusters: 1, codepoints: 5, UnicodeWidthStr::width() == 2 (Read NOTE: https://github.com/unicode-rs/unicode-width)
+
 👁️‍🗨️ Apropriate world
+   ^^^^^^^^^^ highlight here
+
+---
+Face with spiral eyes (U+1F635 U+200D U+1F4AB, Recommended Emoji ZWJ Sequence, v13.1)
+Grapheme clusters: 1, codepoints: 3, UnicodeWidthStr::width() == 4 (Read NOTE: https://github.com/unicode-rs/unicode-width)
+
+😵‍💫 Apropriate world
+     ^^^^^^^^^^ highlight here
+
+---
+Horizontal tab (\t, U+09)
+Grapheme clusters: 1, codepoints: 1, UnicodeWidthStr::width() == 0
+
 	Apropriate world
+^^^^^^^^^^ highlight here
--- a/tests/cmd/stdin-failure-multiwidth.stdout
+++ b/tests/cmd/stdin-failure-multiwidth.stdout
@@ -1,0 +1,30 @@
+error: `Apropriate` should be `Appropriate`
+  --> -:1:2
+  |
+1 | 한 Apropriate world
+  |    ^^^^^^^^^^
+  |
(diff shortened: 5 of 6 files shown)


src/bin/typos-cli/report.rs
```diff
@@ -218,8 +220,11 @@ fn print_long_correction(msg: &Typo, palette: Palette) -> Result<(), std::io::Er
         let line_num = context.line_num.to_string();
         let line_indent: String = itertools::repeat_n(" ", line_num.len()).collect();
 
-        let hl_indent: String = itertools::repeat_n(" ", column).collect();
-        let hl: String = itertools::repeat_n("^", msg.typo.len()).collect();
+        let visible_column = UnicodeWidthStr::width(start.as_ref());
+        let visible_len = UnicodeWidthStr::width(msg.typo);
+
+        let hl_indent: String = itertools::repeat_n(" ", visible_column).collect();
+        let hl: String = itertools::repeat_n("^", visible_len).collect();
```
epage, line 227: > I've intentionally left column variable as is since many IDE counts by grapheme clusters.

I fully recognize I have no experience in this area and am just blindly trying to make things do the right thing  With that said, my couple of concerns
- This feels weird to me that we are rendering columns different than what we are telling people.  is there more where I can read on editors using graphemes and why?
- This feels weird for us to pull in a separate dependency for these two similar roles.

One thing that isn't a problem is the API.  It returns bytes which seems the most correct so people can choose what they want on top of this.

src/bin/typos-cli/report.rs
```diff
@@ -218,8 +220,11 @@ fn print_long_correction(msg: &Typo, palette: Palette) -> Result<(), std::io::Er
         let line_num = context.line_num.to_string();
         let line_indent: String = itertools::repeat_n(" ", line_num.len()).collect();
 
-        let hl_indent: String = itertools::repeat_n(" ", column).collect();
-        let hl: String = itertools::repeat_n("^", msg.typo.len()).collect();
+        let visible_column = UnicodeWidthStr::width(start.as_ref());
+        let visible_len = UnicodeWidthStr::width(msg.typo);
```
epage, line 224: Also, can you add an end-to-end test for this

See https://github.com/crate-ci/typos/blob/master/tests/cli_tests.rs for our snapshot tests

foriequal0:
```rust
fn main() {
    let 한=abc1;
    // The variable is NFC, 1 grpahemes, 1 codepoints.
    // Many Koreans are familiar with this.
    // Rust compiler, VSCode, JetBrains IDE, Notepad++ agree on that 'a' is at column number 11.
    // Most terminal emulators render it as fixed width (2 column width) characters.
    // Old editors used to render it as fixed width too.
    // Many old Korean programmers are accustomed to this.
    // So, Rust compiler puts highlight at column 12 expecting it to be rendered on terminals.
    // 
    // However, many modern editors, and browsers render the variable as variable width characters.
    // So they are slightly misaligned on some environment such as https://play.rust-lang.org/

    let 한=abc2;
    // The variable is NFD. 1 graphemes, 3 codepoints.
    // I expect it should look same and be treated as above.
    // It is an unfamiliar form even for Korean, but It is how macOS handles a filename.
    // It was due to the limitation of HFS+, but filesystem API keep denormalizes filename on APFS which doesn't have the limitation.
    // macOS renders filenames correctly, but Windows renders as decomposed form, which annoys many Koreans.
    //
    // Anyway, all editors fail to handle correctly in different way.
    // - Rust compiler:
    //   It puts highlight at column 12, but says 'a' is at column 13.
    //   Rust playground https://play.rust-lang.org/ that uses https://ace.c9.io/, and it
    //    - displays the variable as 'ㅎㅏㄴㄱㅡㄹ', which is incorrect,
    //    - and puts a cursor in an incorrect position. The cursor is put between decomposed letters,
    //      and the visible cursor position is different with the editing position.
    //    - It requires 3 strokes to delete a grapheme cluster as if it is compositing,
    //      while many Korean IME deletes a NFC character at once after the composition is completed. 
    // - VSCode:
    //   It says 'a' is at column 13.
    //   It correctly displays the variable and puts a cursor in a correct position.
    //   It requires 3 keystrokes to delete.
    // - Notepad++
    //   It says 'a' is at column 13.
    //   It is displayed as 'ㅎㅏㄴㄱㅡㄹ', but in a slightly different form.
    //   It puts a cursor in an incorrect position. The cursor is put between decomposed letters.
    //   It requires 3 keystrokes to delete.
    // - JetBrains IDE
    //   It says 'a' is at column 13.
    //   It correctly displays the variable.
    //   It fails to put a cursor in a correct position. The cursor is put in the middle of the character (at 1/3 of the character, etc.)
    //   It requires 3 strokes to delete.

    let 👁️‍🗨️=abc3;
    // The variable is "U+1F441 U+FE0F U+200D U+1F5E8 U+FE0F". 1 graphemes, 6 codepoints.
    // Also, all editors fail to handle correctly in different way.
    // - Rust compiler:
    //   It puts highlight at column 12, but says 'a' is at column 15.
    //   Rust playground treats the variable as 3 characters '👁️ 🗨️' in the editor, but it is displayed as 2 characters '👁️🗨️' in the compilation result panel.
    //   It requires 3 keystrokes to navigate across the variable with the cursor.
    //   It required 5 keystrokes to delete.
    // - VSCode:
    //   It says 'a' is at column 15.
    //   It correctly displays the variable and puts a cursor in a correct position.
    //   However, deleting the character requires 2 keystrokes.
    // - Notepad++
    //   It says 'a' is at column 15.
    //   It correctly displays the variable.
    //   It requires 5 keystrokes to navigate across the variable with the cursor, and the cursor sits in the middle of the character (1/5 of the character, etc.).
    //   It required 5 keystrokes to delete.
    // - JetBrains IDE
    //   It says 'a' is at column 15.
    //   It displays as 2 characters '👁️🗨️'.
    //   It requires 2/1(ZWJ)/2, keystrokes to navigate across the variable with the cursor, and the cursor sits in the middle of the character (1/2 of the character, etc.).
    //   It required 5 keystrokes to delete.

    let 😵‍💫=abc4;
    // The variable is "U+1F635 U+200D U+1F4AB". 1 graphemes, 3 codepoints.
    // Also, all editors fail to handle correctly in different way.
    // - Rust compiler:
    //   It puts highlight at column 14, but says 'a' is at column 13.
    //   Rust playground treats the variable as 3 characters '😵 💫' in the editor, but it is displayed as 2 characters '😵💫' in the compilation result panel.
    //   It requires 3 keystrokes to navigate across the variable with the cursor.
    //   It required 3 keystrokes to delete.
    // - VSCode:
    //   It says 'a' is at column 13.
    //   It incorrectly displays the variable as '😵💫'.
    //   It requires 1 keystroke to navigate across the variable with the cursor.
    //   It required 2 keystrokes to delete.
    // - Notepad++
    //   It says 'a' is at column 13.
    //   It correctly displays the variable.
    //   It requires 1/1(ZWJ)/1 keystrokes to navigate across the variable with the cursor.
    //   It required 3 keystrokes to delete.
    // - JetBrains IDE
    //   It says 'a' is at column 13.
    //   It displays as 2 characters '😵💫'.
    //   It requires 1/1(ZWJ)/1 keystrokes to navigate across the variable with the cursor.
    //   It required 3 keystrokes to delete.

	abc5
    // The above line starts with a horizontal tab ('\t', 0x09), and I'm using 4 spaces tab.
    // - Rust compiler:
    //   It puts highlight at column 5, but says 'a' is at column 2.
    // - VSCode:
    //   It says 'a' is at column 5.
    // - Notepad++
    //   It says 'a' is at column 5.
    // - JetBrains IDE
    //   It says 'a' is at column 5.
}
```

Rust playground:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6494413102a03a812fd193476c16d2aa
image

Windows Terminal
image

VSCode
image

Notepad++
image

JetBrains Rider
image

Rust compiler output in Rust playground (Firefox, 100)
image

Rust compiler output in Windows Terminal, WSL
image

src/bin/typos-cli/report.rs
foriequal0: I'm just a regular user of text editors 😄 sorry. I couldn't find a good one to read.
However, I can summerize some editor and compiler's behavior as a regular user. See [this comment](https://github.com/crate-ci/typos/pull/465#issuecomment-1109355495).
In short, Rust compiler and editors count column numbers in codepoints, while editors count a horizontal tab character according to indenting configuration. And the codepoints count, the UnicodeWidthStr width, the actual column width in display, keystorke count to navigate, edit, and the grapheme cluster count are incoherent.
I personally like grapheme cluster based width.

epage:
Thanks for taking the time to explain things to me!  Would like t see the cases explained for the tests and after that, its good to merge and release.

foriequal0:
I've added descriptions to test fixtures.
I think column numbers and alignments are unintuitive. I'll cover it on another PR.

epage:
Thanks!

> Should I make a separate PR?

Sure!

PR #501, opened 2022-06-16
feat(action): Report typos as annotations

### LINKED ISSUE #429: New reporter - github action annotation
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-warning-message

essentially
`::error file={name},line={line},endLine={endLine},title={title}::{message}`
ex
`echo "::error file=app.js,line=1,col=5,endColumn=7::Missing semicolon"`

All of the fields pretty optional, but I think it could do file, line and msg pretty easily.

I tried to figure out enough rust to implement this, but failed.

epage:
Inspired by a snippet from @arminrosu on #429.

Fixes #429

### DIFF
--- a/.github/workflows/test-action.yml
+++ b/.github/workflows/test-action.yml
@@ -2,12 +2,36 @@ name: Test GitHub Action
 on: [pull_request]
 
 jobs:
-  run:
+  shallow:
     name: Spell Check with Typos
     runs-on: ubuntu-latest
     steps:
     - name: Checkout Actions Repository
       uses: actions/checkout@v2
+
+    - name: Prepare file with mistakes.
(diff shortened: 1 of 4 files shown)


jplatte:
Could this be the cause for this strange warning / annotation?

Screenshot_2022-06-22_175104

epage:
The "should be" lines up but
- Not sure where "Style" is coming from
- We shouldn't be checking `.png` files
- We shouldn't be correcting `"null"` to `""` unless these values are coming from `jq` instead

Oh, I bet this is for the json message we output for binary files.  We don't filter those out, so we are creating these annotations for them.

Mind creating an issue for this?

PR #533, opened 2022-08-14
fix: remove thead -> thread

uyarn:
to fix issue : https://github.com/crate-ci/typos/issues/532

PR #534, opened 2022-08-15
Correct "opauqe" to "opaque"

ayazhafiz:
I can't find any references to "opauqe" as an actual word, so I believe
this to be safe.

Thanks for this library!


epage:
Thanks!

I'll probably hold off on a release of this until #533 is ready, if thats ok

ayazhafiz:
Of course, no rush

PR #535, opened 2022-08-16
Add "deffer" to typo list

jplatte:


jplatte:
I tried to fix the codegen thing with `SNAPSHOTS=overwrite cargo test --workspace` and then with `SNAPSHOTS=overwrite cargo test -p typos-dict --test codegen`, but it doesn't do anything. The error also doesn't happen locally.

epage:
> I tried to fix the codegen thing with `SNAPSHOTS=overwrite cargo test --workspace` and then with `SNAPSHOTS=overwrite cargo test -p typos-dict --test codegen`, but it doesn't do anything except run a few tests successfully.
> 
> The error seen in the CI job also doesn't happen locally.

Strange, I just checked out your branch and ran `cargo test -p typos-dict --test codegen` and I had a test failure.  I then ran  `SNAPSHOTS=overwrite cargo test -p typos-dict --test codegen` and it succeeded by modifying the source file.

jplatte:
This is what I get:

```
    Finished test [unoptimized + debuginfo] target(s) in 0.04s
     Running tests/codegen.rs (target/debug/deps/codegen-2710dece8aac08d5)

running 1 test
test codegen ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 3.94s
```

jplatte:
I did a `cargo clean` and re-ran, it succeeded again but overwrote that one Rust file this time (even without `SNAPSHOTS` being set).

epage:
> I did a `cargo clean` and re-ran, it succeeded again but overwrote that one Rust file this time (even without `SNAPSHOTS` being set).

We use `include_bytes!`.  I suspect there is a cargo bug in here with detecting rebuilds.  I might try to go and reproduce this.

However, overwriting without SNAPSHOTS should not happen.  All file writing is done via snapshot testing using snapbox which is being heavily used in several of my crates.  I can't see why this would write when you didn't ask for it which is what worries me.

PR #538, opened 2022-08-17
feat(dict): Add typos for "inappropriate[ly]"

jplatte:


jplatte:
Sooo this time `SNAPSHOTS=overwrite` seemed to work, but CI is still unhappy? 😕

epage:
> Sooo this time `SNAPSHOTS=overwrite` seemed to work, but CI is still unhappy? confused

If you ran just the `codegen` "test", then the failure would make sense,  There is also one to check the formatting of the csv file.

jplatte:
I did that first, and got a failure, which I was able to fix with the overwrite mode. But clearly ci wants to make further changes.

jplatte:
Ah it is indeed now the `verify` test that fails, not `codegen`.

PR #596, opened 2022-10-22
fix(dict): suggest cancellation, not cancelation for cancelaltion

scop:
For consistency with what we do e.g. with `cancellato`.

I'm not sure what's up with the codegen, I followed the recipe in CONTRIBUTING.md, but it fails and wants to revert the change to words.csv.

### DIFF
--- a/crates/typos-dict/assets/words.csv
+++ b/crates/typos-dict/assets/words.csv
@@ -7058,7 +7058,7 @@ canadianese,canadians
 canadias,canadians
 cananbis,cannabis
 cancelability,cancellability
-cancelaltion,cancelation
+cancelaltion,cancellation
 cancelas,cancels
 cancelations,cancellations
 canceles,cancels


scop:
Hm, do I understand correctly that the results are kind of intentionally inconsistent with regards to the dialect using out of the box defaults, maybe caused by the selection to the closest dialect chosen for each encountered word? I think a consistent per-run or per-file rather than word by word selection would be an improvement over that.

epage:
Yes, we do not consistently correct to a single dialect by default.  The default is all dialects.  The user can choose a specific dialect if they wish and we will push words towards that.  To that end, we correct to the dialect that most closely matches the typo.

> I think a consistent per-run or per-file rather than word by word selection would be an improvement over that.

Not entirely too sure what you are suggesting.  However, if you have an idea for an alternative mode of operation, I'd recommend we continue discussing that in an issue.

PR #612, opened 2022-11-21
New typos

NathanBaulch:
Great tool! I recently fixed thousands of typos in a [popular open source project](https://github.com/OpenAPITools/openapi-generator) and thought maybe I could contribute the 133 typos your tool missed. Let me know if any of the corrections are inappropriate.

Unfortunately I'm new to Rust and wasn't able to get the `cargo test --workspace` command working, so this PR is a work in progress. I'm sure it's just something environmental. Here's the backtrace:
```
---- codegen stdout ----
thread 'codegen' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: NotFound, message: "program not found" }', crates\codespell-dict\tests\codegen.rs:9:54
stack backtrace:
   0: rust_begin_unwind
             at /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52/library\std\src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52/library\core\src/panicking.rs:142:14
   2: core::result::unwrap_failed
             at /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52/library\core\src/result.rs:1814:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52\library\core\src/result.rs:1107:23
   4: codegen::codegen
             at .\tests\codegen.rs:9:19
   5: codegen::codegen::{{closure}}
             at .\tests\codegen.rs:4:1
   6: core::ops::function::FnOnce::call_once
             at /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52\library\core\src\ops/function.rs:248:5
   7: core::ops::function::FnOnce::call_once
             at /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52/library\core\src\ops/function.rs:248:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

PR #613, opened 2022-11-22
feat(dict): 133 assorted typos

NathanBaulch:
Recreating #612...

Great tool! I recently fixed thousands of typos in a [popular open source project](https://github.com/OpenAPITools/openapi-generator) and thought maybe I could contribute the 133 typos your tool missed. Let me know if any of the corrections are inappropriate.


NathanBaulch:
Thanks for cleaning that up, next time will be smoother. 👍

crates/typos-dict/assets/words.csv
```diff
@@ -32948,6 +33022,7 @@ nigthmares,nightmares
 nihilim,nihilism
 nihilisim,nihilism
 nihilsim,nihilism
+nilable,nillable
```
Blacksmoke16, line 33025: Not sure the best way to go about this, but `nilable` is actually fairly common in the Ruby world. E.g. https://sorbet.org/docs/nilable-types.

I worked around it by just adding it to `extended-words`, which is fine if we think this is the preferred spelling outside of certain ecosystems, but it could be a point of false positives.

epage: I'd rather allow more typos than correct things that shouldn't be typos.

@Blacksmoke16 would it be ok for `nilable` to be in the default `extend-words` for Ruby only or would it appear in enough other contexts to need to be generally allowed?

Blacksmoke16: > but am also curious about how best to handle common spelling mistakes that are widely accepted in certain languages/frameworks/communities

IMO it would be nice if there was a more selective way to denote a typo is expected. E.g. such that I could mark a specific usage of `referer` as correct, but not prevent that from flagging other contexts. Currently, afaik, the only way to handle that is to just globally accept the typo by adding it to `extended-words`, or accept it for a given "type" of file.

> would it be ok for `nilable` to be in the default `extend-words` for Ruby only

For my particular use case, it would need to be Ruby (.rb), Crystal (.cr), and Markdown (.md) in order to cover usages and documentation related to the code. Did a quick google of `"nilable"` and saw some things in there about some Go libraries, and ClojureScript. 
 So guess it depends on how many contexts is enough to suggest it should be globally allowed.

epage: Created #635 for nilable.

For referer, we have a couple of issues around narrowing things.  If one doesn't fit, feel free to create a new issue.

PR #636, opened 2022-12-08
update cases I encountered

alekspickle:


epage:
Looks like there are some things to work through from running `cargo test -p typos-dict`

alekspickle:
welp, seems it's not needed, too many conflicts now

PR #647, opened 2023-01-03
git ignore VS Code config folder

cbachhuber:
VS Code is widely used and creates a `.vscode/` folder in the repo for project-specific settings. I think we should ignore that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
+.vscode/
 target


.gitignore
```diff
@@ -1 +1,2 @@
+.vscode/
```
epage, line 1: I'd prefer people ignore their editor of choice in their per-user gitignores.

PR #654, opened 2023-01-17
Adds 'regylar' as a typo for 'regular'.

dupgit:
Hi,

I propose to add this typo ('regylar' for 'regular') in the patch included. I'm not very used to git and not sure that this is the right process to submit a PR. Please tell me if I should do things differently.

Regards,

Olivier.


epage:
Thanks

PR #657, opened 2023-01-26
Allow writing changes in the github action

bnjbvr:
Hi! I think it'd make sense to allow writing changes in the github action itself, so that this can be combined with other actions like https://github.com/getsentry/action-git-diff-suggestions. That way, I can run `typos` in my repository, and instead of a hard failure, it will suggest code changes to fix the typos.

See https://github.com/bnjbvr/octotest/pull/5 for an example.

(The entrypoint file will run the command twice, so I suspect that it will only have an effect the first time. I can tweak the script so it doesn't, but as far as I could tell, this didn't add extra warnings or other issues, so that seemed ok to me. Let me know if you'd like me to fix that!)

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -20,6 +20,11 @@ inputs:
     required: false
     default: false
 
+  write_changes:
+    description: "Write changes to the repository"
+    required: false
+    default: false
+
   config:
     description: "Use a custom config file."
     required: false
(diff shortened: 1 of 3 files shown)


bnjbvr:
Interesting, I didn't know about Sarif! Actions like the one I've combined this one with could probably read from sarif outputs in the future, and then emit code suggestions based on that. For now, the current solution has been working great for us, for the past week.

I've now added documentation; can you take another look, please? Thanks!

epage:
Thanks!

PR #667, opened 2023-02-24
Edited README to show install instructions for `conda`

### LINKED ISSUE #666: Informative: `typos` is now added to conda-forge
I want to let you know that I added `typos` to conda-forge. It has Windows, Linux, Mac OSX (including arm64) support. This may be useful to `conda install typos` in a system when `cargo` or `brew` are unavailable.

Relevant feedstock: https://github.com/conda-forge/typos-feedstock

There was a use-case at work where I needed to access `typos` through conda-forge, since that was the only source the firewall would allow packages to come from. That's an aside. But it may be useful to others who simply do not want to use `cargo`.

This is not an issue, rather simply informational.

DamianBarabonkovQC:
Fixes #666 

Just as the title says.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,11 @@ Or use [Homebrew](https://brew.sh/) to install:
 ```bash
 brew install typos-cli
 ```
+
+Or use [Conda](https://conda.io/) to install:
+```bash
+conda install typos
+```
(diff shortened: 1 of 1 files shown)


epage:
Thanks!

PR #676, opened 2023-03-09
feat(dict): Add empheral -> ephemeral

jplatte:


epage:
Something I've been considering is providing a place for people to put one off corrections that they aren't pressed to get merged and I occasionally go in an batch add them

Thoughts?

jplatte:
Sounds good to me.

Issue #677, opened 2023-03-10
Non-critical corrections - March 2023 Edition


jplatte:
Please list any corrections you noticed `typos` doesn't fix and we'll roll these up into a single PR, making it lighterweight for adding new corrections.

jplatte:
- `empheral` -> `ephemeral` (#676)
- `referenecs` -> `references`

epage:
**Meta**: We can try for a per-month issue for this for now.  My original thought had been to use a Discussion with the current "correct answer" being the current month.  This would provide a stable link for places like the README to use.  The downside is its too easy for people to reply in the wrong thread / create a new thread and Github doesn't have the thread-editing tools of Zulip afaik

epage:
- `configurating` -> `configuring`
- `configurating` -> `technical`
- `signficiantly` -> `significantly`
- `substitues` -> `substitutes`

Source: rust-lang/cargo#11794 (found after rust-lang/cargo11561)

jplatte:
**Meta:** I think the nice things about an issue are

- you can pin it (though maybe this is a thing for discussions too)
- people are more used to issues and go there first

But I think that's it really, mostly I did it this way because that's what came to mind for me. I'll be happy to suggest new corrections elsewhere if you decide to try sth. else at any point.

adamchainz:
`encryptation` -> `encryption`

cmdcolin:
deature->feature

PR #679, opened 2023-03-12
feat(dict): Add: erronerous -> erroneous

phip1611:
This adds a typo variant that I found in the code base of the [uefi](https://github.com/rust-osdev/uefi-rs) crate.


epage:
FYI we now have #677 for people collecting non-critical corrections to add them at once

PR #687, opened 2023-03-14
feat(ci): Use composite to run action

### LINKED ISSUE #678: Github Action can sometimes be slow do to building of the image
Currently using Docker to run GitHub Action, its required to building the image every time so it's too slow. Run it using Node.js to make sure it's faster.
We can use https://github.com/ad-m/github-push-action/blob/fe38f0a751bf9149f0270cc1fe20bf9156854365/start.js to execute a Shell script.

Action using node: https://github.com/WaterLemons2k/typos/actions/runs/4390226555

I would like to open a PR for this issue if possible.

**I don't know much about Rust and Node.js, so please excuse my mistakes.**

WaterLemons2k:
Closes #678.

Using composite instead of docker to avoid building image. If the `typos` command doesn't exist, download and extract it.

Also, reference https://github.com/crate-ci/typos/pull/267#discussion_r642705684 adds a pre-release replacement for entrypoint.sh.

Action using composite: https://github.com/WaterLemons2k/typos/actions/runs/4411614891


action.yml
```diff
@@ -30,8 +30,18 @@ inputs:
     required: false
 
 runs:
-  using: "docker"
-  image: "docker/Dockerfile"
+  using: 'composite'
+  steps:
+    - id: typos
+      run: $GITHUB_ACTION_PATH/docker/entrypoint.sh
```
epage, line 36: I feel like this file should move into an `action` folder or something

crates/typos-cli/Cargo.toml
```diff
@@ -26,6 +26,7 @@ pre-release-replacements = [
   {file="../../CHANGELOG.md", search="<!-- next-header -->", replace="<!-- next-header -->\n## [Unreleased] - ReleaseDate\n", exactly=1},
   {file="../../CHANGELOG.md", search="<!-- next-url -->", replace="<!-- next-url -->\n[Unreleased]: https://github.com/crate-ci/typos/compare/{{tag_name}}...HEAD", exactly=1},
   {file="../../docker/Dockerfile", search="ARG VERSION=.*", replace="ARG VERSION={{version}}", min=1},
```
epage, line 28: Since this Dockerfile is only for the action, we should remove it and the replacement for it

docker/entrypoint.sh
```diff
@@ -2,6 +2,14 @@
 
 set -eu
 
+# If typos does not exist, download and extract it
+if [[ ! -f /usr/local/bin/typos ]]; then
```
epage, line 6: Why is this needed?  It caches the location we download to?

If so, when does it invalidate? If we have a new release will users be sure to get the new version?

action.yml
WaterLemons2k: Moved.

crates/typos-cli/Cargo.toml
WaterLemons2k: Fixed.

docker/entrypoint.sh
WaterLemons2k: It doesn't caches anything. It was initially added because if you don't download typos within docker, you need to download it within the shell.

However, later it was discovered that there was already an if statement checking if the `typos` command exists, so they were merged together.

WaterLemons2k:
Capitalized the first letter of the title and merged the commit.

epage:
Thanks!

PR #708, opened 2023-04-06
feat: Add `--file-list <PATH>` to choose files to check

### LINKED ISSUE #707: Feature request: pass file list through stdin
I have a long list of files (newline or `\0` separated) to run through `typos`, and I don't see a good way to pass them other than a bash loop executing `typos` one by one (which is a waste of time).

Passing like:

```sh
typos $files
```

have issues around globbing and command line argument limit, and can't handle `\0`.

Would be nice is if I could:

```sh
echo "$files" | typos --stdin-paths   # for newline separated paths
echo "$files0" | typos --stdin-paths0 # for zero-separated paths
```

dpc:
Fix #707

Just the draft as I had some time. Please feel free to take over if you'd like, I'm not sure if and when I'll find to finish.


crates/typos-cli/src/bin/typos-cli/main.rs
```diff
@@ -269,6 +286,46 @@ fn run_checks(args: &args::Args) -> proc_exit::ExitResult {
...
+                    (paths, cur_path)
+                }
+            },
+        )
+        .0)
+}
+
+#[cfg(not(target_family = "unix"))]
+fn unix_read_paths_from_stdin() -> Result<Vec<PathBuf>, proc_exit::Exit> {
+    Err(proc_exit::sysexits::CONFIG_ERR)
```
dpc, line 326: Better error message?

crates/typos-cli/src/bin/typos-cli/args.rs
```diff
@@ -50,6 +50,14 @@ pub(crate) struct Args {
     #[arg(default_value = ".")]
     pub(crate) path: Vec<std::path::PathBuf>,
 
+    /// Read the list of newline separated paths from stdin
+    #[arg(long, conflicts_with = "stdin_paths0")]
+    pub(crate) stdin_paths: bool,
+
+    /// Read the list of '\0' separated paths from stdin
+    #[arg(long, conflicts_with = "stdin_paths")]
+    pub(crate) stdin_paths0: bool,
```
dpc, line 59: Possible to easily hide this on windows?

crates/typos-cli/src/bin/typos-cli/args.rs
```diff
@@ -50,6 +50,14 @@ pub(crate) struct Args {
     #[arg(default_value = ".")]
     pub(crate) path: Vec<std::path::PathBuf>,
 
+    /// Read the list of newline separated paths from stdin
```
epage, line 53: Placeholder for any design issues to first be resolved in #707

crates/typos-cli/src/bin/typos-cli/main.rs
```diff
@@ -156,8 +159,26 @@ fn run_checks(args: &args::Args) -> proc_exit::ExitResult {
...
+    } else if args.stdin_paths0 {
+        Some(unix_read_paths_from_stdin()?)
+    } else {
+        None
+    };
+
+    for path in stdin_paths.as_ref().unwrap_or(&args.path) {
         let cwd = if path == std::path::Path::new("-") {
+            if args.is_using_stdin_paths() {
+                proc_exit::sysexits::NO_INPUT.ok()?;
```
epage, line 180: This will likely need an error message to go along with a failing exit code

crates/typos-cli/src/bin/typos-cli/main.rs
```diff
@@ -156,8 +159,26 @@ fn run_checks(args: &args::Args) -> proc_exit::ExitResult {
...
+                .collect::<Result<_, _>>()
+                .with_code(proc_exit::sysexits::IO_ERR)?,
+        )
+    } else if args.stdin_paths0 {
+        Some(unix_read_paths_from_stdin()?)
+    } else {
+        None
+    };
+
+    for path in stdin_paths.as_ref().unwrap_or(&args.path) {
```
epage, line 177: Since the stdin paths don't conflict with user provided paths in clap, this would be ignoring user data

If you decide to make them all mutually exclusive, you can add a `group = "source"` to each argument.  By default, clap treats groups as mutually exclusive

crates/typos-cli/src/bin/typos-cli/main.rs
```diff
@@ -269,6 +290,39 @@ fn run_checks(args: &args::Args) -> proc_exit::ExitResult {
...
+        paths.push(PathBuf::from(OsString::from_vec(rest)));
+    }
+    Ok(paths)
+}
+
+#[cfg(not(target_family = "unix"))]
+fn unix_read_paths_from_stdin() -> Result<Vec<PathBuf>, proc_exit::Exit> {
+    proc_exit::sysexits::CONFIG_ERR.ok()?;
+}
+
```
epage, line 325: Options
- We keep this platform-specific
- We use `os_str_bytes`
- We wait on rust-lang/rust#109698
- We guiltily act as if rust-lang/rust#109698 is reality and transmute

epage: - We punt on null terminated paths for now

crates/typos-cli/src/bin/typos-cli/main.rs
dpc: I don't know how to add error messages. :D Please help.

crates/typos-cli/src/bin/typos-cli/main.rs
dpc: TIL.

dpc: The biggest problem is the default value (`.`). If it wasn't for that we could just use `.chain` and use both stdin-paths and cli passed onces. 

Though for simplicity sake, it's probably best to forbid using both.

crates/typos-cli/src/bin/typos-cli/main.rs
dpc: I don't like waiting. If we make it platform-specific now, some good brave soul can make it works on Windows once they need it.

`os_str_bytes` seems fine with me if you're OK with an extra dependency.

epage: `os_str_bytes` is already a dependency through clap (until we upgrade; it was dropped recently)

crates/typos-cli/src/bin/typos-cli/main.rs
epage: `proc_exit::sysexits::NO_INPUT.with_message(...)`

dpc: TIL. But I decided that when passing through stdin `-` should be treated like any other path.

crates/typos-cli/src/bin/typos-cli/main.rs
dpc: So it seems like it's not wanted?

I don't really have much time to spend on this and I don't care about Windows, so I would rather leave as is. It doesn't make anything worse for Windows users, AFAICT.

crates/typos-cli/src/bin/typos-cli/main.rs
```diff
@@ -156,8 +159,24 @@ fn run_checks(args: &args::Args) -> proc_exit::ExitResult {
...
+                .collect::<Result<_, _>>()
+                .with_code(proc_exit::sysexits::IO_ERR)?,
+        )
+    } else if args.stdin_paths0 {
+        Some(unix_read_paths_from_stdin()?)
+    } else {
+        None
+    };
+
+    for path in stdin_paths.as_ref().unwrap_or(&args.path) {
```
epage, line 177: I think I'd like a comment saying that the reason we can do `unwrap_or` is because these two sources are mutually exclusive

crates/typos-cli/src/bin/typos-cli/main.rs
```diff
@@ -1,4 +1,7 @@
-use std::io::Write;
+use std::{
+    io::{self, Read, Write},
+    path::PathBuf,
+};
```
epage, line 4: style nit: I prefer
- Not using multi-line `use`s
- Minimizing the user of `use` generally (traits with `as _` being an exception)
  - Of these, the one I probably care most about is `use std::self` as that doesn't really buy us all that much

crates/typos-cli/src/bin/typos-cli/main.rs
```diff
@@ -156,8 +159,24 @@ fn run_checks(args: &args::Args) -> proc_exit::ExitResult {
...
+        )
+    } else if args.stdin_paths0 {
+        Some(unix_read_paths_from_stdin()?)
+    } else {
+        None
+    };
+
+    for path in stdin_paths.as_ref().unwrap_or(&args.path) {
+        // Note paths are passed through stdin, `-` is treated like a normal path
+        let cwd = if !args.is_using_stdin_paths() && path == std::path::Path::new("-") {
```
epage, line 179: Deeper in, we'll still treat `-` as `stdin`, so this specialization right here doesn't quite make sense

dpc: Really? That doesn't seem ideal. What if someone wanted to lint a codebase that happens to have `-` somewhere?

epage: If that is a concern, then solving it is independent of this PR anyways and we should start with an issue from soneone running into this in the wild.

crates/typos-cli/src/bin/typos-cli/main.rs
epage: Its been a while since I last looked at it and everything was unstable then...

epage: Looks like `imports_granularity` is not stable: https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#imports_granularity

crates/typos-cli/src/bin/typos-cli/main.rs
dpc: https://docs.rs/os_str_bytes/latest/os_str_bytes/struct.EncodingError.html

> [Result::unwrap](https://doc.rust-lang.org/nightly/core/result/enum.Result.html#method.unwrap) should almost always be called on results containing this error. It should be known whether or not byte sequences are properly encoded for the platform, since [the module-level documentation](https://docs.rs/os_str_bytes/latest/os_str_bytes/index.html#encoding) discourages using encoded bytes in interchange. Results are returned primarily to make panicking behavior explicit.

This sounds like it is not supposed to be used on user provided data (the whole crate)? When I tried to use it anyway, it complained about having to set some env variable.

Anyway. I'm out of time and I only go out of my way to support commercial software when paid to do it. Its fine if you don't want to land it.

crates/typos-cli/src/bin/typos-cli/main.rs
dpc: Reverted, how it's OK.

dpc: (BUT :D . I just have to complain about it: these magic strings like `-` are a terrible idea and a bane of CLI software. They were excusable 20 years ago, when they had 1M of RAM and had to write software in C. There's no good reason to do it this way in Rust, where it could be `--stdin` or something like that. No disrespect. Just a pet peeve.)

dpc:
Eh. I still want it. How about now?

dpc:
Ehh. I don't know how to properly handle this on Windows. I don't know the platform, and it seems it might actually require checking the current system encoding and some other stuff like that?

It's ridiculous. Actual users could have been enjoying this feature for months now, and we stuck trying to support all these 0.7 users of `typos` on Windows, none of which even know what 0-separated strings are. **Not supporting it on Windows does not break any existing feature for Windows users**.

Here is my solution for Windows users who need it: either submit a PR, or here is $0 needed to install an operating system for real work: :dollar: .

dpc:
I have rebased on top of #837 which fixes one pain point of this PR I was hitting.

Would be nice if we could land. It's a really useful feature. Have been using it from my fork all this time. If there's anything other than obsolete OSes support, I'd be happy to address it. I would really fix it even for Windows if I could. I really have better thing to do than annoy you about it. :rofl:

crates/typos-cli/tests/cli_tests.rs
```diff
@@ -1,5 +1,7 @@
 #[test]
 #[cfg(feature = "dict")]
 fn cli_tests() {
-    trycmd::TestCases::new().case("tests/cmd/*.toml");
+    trycmd::TestCases::new()
+        .case("tests/cmd/*.toml")
+        .case("tests/cmd/*.trycmd");
```
epage, line 6: Is this change needed?

crates/typos-cli/tests/cmd/stdin-paths0.in/_typos.toml
```diff
@@ -0,0 +1,8 @@
+[files]
```
epage, line 1: Whats the role of these files?

crates/typos-cli/tests/cmd/stdin-paths.in/_typos.toml
```diff
@@ -0,0 +1,8 @@
+[files]
```
epage, line 1: Could you rename `stdin-paths` to `file-list-stdin`?

epage:
I believe the comments I just made are the extent of what is needed and I can merge after this.

(nit: this would be a "feat" type, rather than a "chore" though this PR process might have felt like one...)

dpc:
> (nit: this would be a "feat" type, rather than a "chore" though this PR process might have felt like one...)

Features are chores too.

crates/typos-cli/tests/cmd/stdin-paths0.in/_typos.toml
dpc: They're an important leftover I forgot about. :D

crates/typos-cli/tests/cli_tests.rs
dpc: Wasn't me.

dpc:
@epage done

epage (approved):

epage:
> Features are chores too.

I'm curious, how do you define the "chore" type?  I'm not seeing any more formal definition.  I tend to use it for non-user-impacting project bookkeeping (e.g. changing CI pipelines).  Or from another perspective, I view it as "this doesn't go in a changelog" while a `feat` and a `fix` would.

dpc:
> I'm curious, how do you define the "chore" type? I'm not seeing any more formal definition. I tend to use it for non-user-impacting project bookkeeping (e.g. changing CI pipelines). Or from another perspective, I view it as "this doesn't go in a changelog" while a `feat` and a `fix` would.

I suck at conventional commits and I'm not even sure why I'm trying...

> I tend to use it ...

Makese sense. I should try to remember it. "Doesn't go in a changelog"...

PR #714, opened 2023-04-22
fix(action) Make wget progress indicator much smaller

szepeviktor:
Because the action runs without a tty, `wget` defaults to `--progress dot`.  This prints a dot per 1K of data.  Due to the size of `typos`, switching to `--progress dot:mega` is better as it only prints one dot per 64K.

```
     0K ........ ........ ........ ........ ........ ........ 51% 11.7M 0s
  3072K ........ ........ ........ ........ ........ .....   100% 12.0M=0.5s
```

### DIFF
--- a/action/entrypoint.sh
+++ b/action/entrypoint.sh
@@ -32,7 +32,7 @@ fi
 if [[ -z $(which ${CMD_NAME} 2>/dev/null) ]]; then
     VERSION=1.14.8
     log "Downloading 'typos' v${VERSION}"
-    wget https://github.com/crate-ci/typos/releases/download/v${VERSION}/typos-v${VERSION}-x86_64-unknown-linux-musl.tar.gz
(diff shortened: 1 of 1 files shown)


epage:
In the PR, could you contrast it with what it is replacing and explain why this is needed?

From my guess, when the action is run,  `--progress dot` is being used instead of `--progress bar` because it isn't a TTY and the tarball is big enough that a larger increment is needed for printing out a dot.  It is a big help to maintainers to include this kind of context so they can more easily review and merge changes than putting it on them to investigate and guess what is happening.

(honestly, never really noticed the extra dots because I never end up looking at the job output)

szepeviktor:
> never really noticed the extra dots

All I do is under codename "working" is looking at bytes.
Underlying technologies.

szepeviktor:
> explain why this is needed

Excuse me. I handled this repository as my own.
Actually I have a basic rule for bug reports:

1. Where is it happening?
2. Current state
3. Desired state

szepeviktor:
Thank you.
Please squash my commits.

szepeviktor:
Thank you.

Issue #727, opened 2023-05-21
RFE: ignore build system input `.in` extension when determining the type


scop:
Would be good to strip known template/backup filename extensions before attempting a match against a filename/extension glob.

Here's an implementation of that I did some time ago for `bat`: https://github.com/sharkdp/bat/pull/1687

epage:
Template files are a weird case because they contain multiple languages.  Which language should we check?

For backup files, I assume we would instead want to ignore all of them.

scop:
Right, ignoring backups would make sense for typos.

Regarding templates, the case I'm mostly interested in is the `.in` generic build system input one; for that I think the right thing to do is to strip the suffix and decide based on what's left in the name after that. I don't remember coming across one that would actually have multiple languages in it, but just some placeholders to be replaced.

For this specific case, using appropriate type specific extend-globs like
```toml
[type.sh]
extend-glob = ["*.sh.in"]
```
...does seem to serve as a workaround for the time being.

epage:
> I don't remember coming across one that would actually have multiple languages in it, but just some placeholders to be replaced.

I was referring to when you have a template language mixed in with the generated output.

You could say php/html is an example of this.  See also liquid, tera, ninja, handlebars, etc.

epage:
Another aspect of backups is that hopefully they are gitignored (in the per-user ignore file) which typos will then ignore by defalt.

scop:
Right, agreed on both, and the PHP example is an interesting one.

Anyway as far as I'm concerned, this RFE reduces to just stripping the literal trailling `.in` suffix (recursively, as at least `*.in.in` are a pattern in use in some cases too) before attempting filename based type finding.

epage:
I'd be fine with just stripping `.in`

PR #729, opened 2023-05-24
feat(cli): Strip `.in` suffix(es) before attempting filename match

scop:
`.in` is typically used for build system template input files, containing some placeholders to replace. In some cases, multiple rounds of replacements are used, each with their own `.in`, so remove all trailing instances of it.

Closes https://github.com/crate-ci/typos/issues/727

### DIFF
--- a/crates/typos-cli/src/file_type.rs
+++ b/crates/typos-cli/src/file_type.rs
@@ -1,4 +1,5 @@
 use std::collections::BTreeMap;
+use std::path::Path;
 
 use kstring::KString;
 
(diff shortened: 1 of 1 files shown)


crates/typos-cli/src/file_type.rs
```diff
@@ -107,7 +107,7 @@ impl Types {
     }
 
     pub fn file_matched(&self, path: &std::path::Path) -> Option<&str> {
-        let file_name = path.file_name()?;
+        let file_name = path.file_name()?.to_str()?.trim_end_matches(".in");
         let mut matches = self.matches.get_or_default().borrow_mut();
         self.set.matches_into(file_name, &mut *matches);
```
epage, line 112: Should we loop over this, checking for matches, in case a user wanted to define a file type for `.in` files?

scop: Good point. I think 2f87caf436c56be10407bdced138b698874b8c79 addresses this.

crates/typos-cli/src/file_type.rs
```diff
@@ -107,9 +107,15 @@ impl Types {
     }
 
     pub fn file_matched(&self, path: &std::path::Path) -> Option<&str> {
-        let file_name = path.file_name()?.to_str()?.trim_end_matches(".in");
+        let mut file_name = path.file_name()?.to_str()?;
```
epage, line 110: Overlooked this before: this will cause us to fail to lookup the type if the file name isn't valid UTF-8 (`.to_str()?`)

scop: Maybe e12b9a49a5fa70234376b3224755d192f2ef964a?

crates/typos-cli/src/file_type.rs
```diff
@@ -107,14 +108,23 @@ impl Types {
...
+                None => break,
+                Some(ext) => {
+                    if ext == "in" {
+                        pbuf.set_extension("");
+                        continue;
+                    }
+                }
+            }
+            break;
         }
```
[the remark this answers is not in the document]
scop: Changed to that in 867da32b1cea6feb3c411b09e639a43ec4f1dfb9.

scop:
Ok, I think this is ready for another look, checks pass now. I didn't rebase against master though, but it merges cleanly.

PR #735, opened 2023-06-02
Introduce EditorConfig

szepeviktor:
@epage As you can see this repository is a home of diversity :)

I leave it up to you. (while encouraging 4 space indents everywhere - even in YAML!)

Tool used: https://github.com/greut/eclint

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,48 @@
+# https://EditorConfig.org
+
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = lf
+charset = utf-8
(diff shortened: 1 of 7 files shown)


epage:
I'm leaning against this.  This only helps with editors with direct support for it, isn't enforced in CI, and duplicates what rustfmt does.

szepeviktor:
There is an eclint action I use in all my workflows.
https://github.com/szepeviktor/debian-server-tools/blob/45815d59350ea6cfd93617e3a589ccc942182aaf/webserver/Continuous-integration-Continuous-delivery.yml#L174-L179

Issue #733, opened 2023-06-01
Non-critical corrections - May 2023 Edition


epage:
Please list any corrections you noticed `typos` doesn't fix and we'll roll these up into a single PR, making it lighterweight for adding new corrections.

Jeremiah-England:
Probably safe to include:
```
acculumating -> accumulating
anualized -> annualized
assset -> asset
avialible -> available
baises -> biases
cliped -> clipped
curature -> curvature
depdencencies -> dependencies
eveything -> everything
formuala -> formula
hense -> hence
pararm -> param
segmnet -> segment
termonology -> terminology
tolearnce -> tolerance
tonange -> tonnage
traids -> triads
verticla -> vertical
```

Might have other matches?

```
tupes -> tuples
poin -> point
```

Jeremiah-England:
collater -> collator

Jeremiah-England:
creat -> create

SOF3:
tpos -> typos (maybe also others like `tops`)

hofbi:
`coachig` -> `coaching`

vinchona:
```
unalbe -> unable
```

jplatte:
onw -> now

it's already recognized as a typo, but the only suggestion is `own`

jrosdahl:
```
buildning -> building
detectionn -> detection
necassary -> necessary
oputput -> output
ove -> oven, over
polpulate -> populate
poulate -> populate
preocessing -> processing, preprocessing
prediciotn -> prediction
requiremenht -> requirement
vecotor -> vector
```

SOF3:
```
implementationis -> implementations

**Comment by jplatte** [comment]
latst -> latest

**Comment by SOF3** [comment]
propgating -> propagating

**Comment by christoff-linde** [comment]
```
acumalated -> accumulated
developoment -> development
incomatible -> incompatible
tesing -> testing
recalcelated -> recalculated
```

**Comment by jrosdahl** [comment]
fromatted -> formatted

**Comment by not-my-profile** [comment]
unsable -> unable

**Comment by not-my-profile** [comment]
pentalty -> penalty

**Comment by szepeviktor** [comment]
```
wordpresss -> wordpress
```

not-my-profile:
Yeah that correction is already there ... I didn't bother to include suggestions already present.

nicholasbishop:
whule -> while, whole

WaterLemons2k:
nor -> or

[...]

PR #736, opened 2023-06-06
Add resutls to results mapping to words.csv

hofbi:


epage:
FYI we also have #733 for doing these in bulk in case that interests you for the future

hofbi:
Ah, thank you, will put it there

PR #747, opened 2023-06-19
fix(dict): Don't correct "accreting"

tklauser:
"accreting" is the present participle of "accrete", see https://en.wiktionary.org/wiki/accreting. Don't correct it to "accrediting".

### DIFF
--- a/crates/typos-dict/assets/allowed.csv
+++ b/crates/typos-dict/assets/allowed.csv
@@ -5,3 +5,4 @@ referer,http header field
 deques,noun
 dequeues,verb
 ons,so `add-ons` works
+accreting,verb of accrete
--- a/crates/typos-dict/assets/words.csv
+++ b/crates/typos-dict/assets/words.csv
@@ -687,7 +687,6 @@ accredidation,accreditation
 accreditied,accredited
 accreditted,accredited
 accress,access
-accreting,accrediting
 accroding,according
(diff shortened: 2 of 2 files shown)


epage:
This was added in #315 with no comment about where it came from

crates/typos-dict/assets/words.csv
```diff
@@ -687,7 +687,6 @@ accredidation,accreditation
 accreditied,accredited
 accreditted,accredited
 accress,access
-accreting,accrediting
```
epage, line 690: I worry about this being added again in the future.  Could you add this to https://github.com/crate-ci/typos/blob/master/crates/typos-dict/assets/allowed.csv so we prevent this from happening again?

tklauser: Thanks for pointing this out. Now added to `allowed.csv`.

crates/typos-dict/assets/allowed.csv
```diff
@@ -5,3 +5,4 @@ referer,http header field
 deques,noun
 dequeues,verb
 ons,so `add-ons` works
+accreting,verb
```
epage, line 8: Could you provide more detail, like "verb of accrete"?

While I was similarly brief with `deques` and `dequeues`, that was more in context of each other (before we corrected one to the other)

PR #751, opened 2023-06-21
Allow some valid words

not-my-profile:
Stop correcting the following valid words:

* simulative
* pervious
* perviously
* intension
* intensional
* intensionally
* unuseful

### DIFF
--- a/crates/typos-dict/assets/allowed.csv
+++ b/crates/typos-dict/assets/allowed.csv
@@ -8,3 +8,5 @@ ons,so `add-ons` works
 accreting,verb of accrete
 zar,currency code for the South African rand
 simulative,adjective
+pervious,adjective
+perviously,adverb
--- a/crates/typos-dict/assets/allowed.csv
+++ b/crates/typos-dict/assets/allowed.csv
@@ -10,3 +10,6 @@ zar,currency code for the South African rand
 simulative,adjective
 pervious,adjective
 perviously,adverb
+intension,noun
+intensional,adjective
+intensionally,adverb
--- a/crates/typos-dict/assets/allowed.csv
+++ b/crates/typos-dict/assets/allowed.csv
@@ -7,3 +7,4 @@ dequeues,verb
 ons,so `add-ons` works
 accreting,verb of accrete
 zar,currency code for the South African rand
+simulative,adjective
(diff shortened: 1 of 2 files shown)


crates/typos-dict/assets/allowed.csv
```diff
@@ -6,3 +6,10 @@ deques,noun
 dequeues,verb
 ons,so `add-ons` works
 accreting,verb of accrete
+sortings,plural of sorting
```
epage, line 9: Can you provide a source for this?

crates/typos-dict/assets/allowed.csv
```diff
@@ -6,3 +6,10 @@ deques,noun
...
+simulative,adjective
+pervious,adjective
+perviously,adverb
```
epage, line 12: Source: https://www.dictionary.com/browse/pervious

crates/typos-dict/assets/allowed.csv
not-my-profile: Merriam-Webster describes the 5th meaning of the "sorting" noun as:

> an instance of sorting

Multiple instances of sorting would thus be "sortings". See [Google Books for examples of "sortings"](https://www.google.com/search?q=%22sortings%22&tbm=bks) (currently yields 22.600 results). The English Wiktionary also has [a page](https://en.wiktionary.org/wiki/sortings) (though that probably doesn't mean much).

not-my-profile: Ah, right ... good catch! I'll split this off.

not-my-profile: I filed #752 for the further discussion on "sortings".

not-my-profile:
Removed `unuseful,useless` since unuseful is a proper word [according to Merriam Webster](https://www.merriam-webster.com/dictionary/unuseful) and substituting one synonym for another seems awfully opinionated.

not-my-profile:
Converted to a draft because the tests currently fail because of:

```
intensionaly,intentionally,intensionally
```

(the order isn't preserved because of #753 ... once #765 has been merged, I'll rebase this PR)

szepeviktor:
@epage Oh no! People type `pervious` every day.
How to configure Typos to keep spotting this typo?

PR #764, opened 2023-06-24
Refactor generate and add tests

not-my-profile:
A preliminary PR for #765.

### DIFF
--- a/crates/typos-dict/tests/verify.rs
+++ b/crates/typos-dict/tests/verify.rs
@@ -145,6 +145,44 @@ fn process<S: Into<String>>(
         .collect()
 }
 
+#[test]
+fn test_merge_duplicates() {
+    assert_eq!(
+        process([("foo", ["bar"]), ("foo", ["baz"])]),
+        dict_from_iter([("foo", ["bar", "baz"])])
+    );
+}
+
+#[test]
+fn test_duplicate_correction_removal() {
+    let dict = process([("foo", ["bar", "bar"])]);
+    assert_eq!(dict, dict_from_iter([("foo", ["bar"])]));
+}
+
+#[test]
+fn test_cycle_removal() {
+    assert!(process([("foo", ["bar"]), ("bar", ["foo"])]).is_empty());
+}
+
+#[test]
+fn test_varcon_removal() {
+    assert!(process([("colour", ["color"])]).is_empty());
+}
+
+#[test]
+fn test_varcon_best_match() {
+    assert_eq!(
+        process([(
+            "neighourhood", // note the missing 'b'
+            ["neighborhood"],
+        )]),
+        dict_from_iter([(
+            "neighourhood",
+            ["neighbourhood"] // note that 'bor' has become 'bour' to match the typo
+        )])
+    );
+}
+
 fn is_word(word: &str) -> bool {
     word.chars().all(|c| c.is_alphabetic())
 }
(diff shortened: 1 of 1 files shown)


crates/typos-dict/tests/verify.rs
```diff
@@ -19,26 +19,21 @@ fn verify() {
...
             let mut typo = i.next().expect("typo").to_owned();
             typo.make_ascii_lowercase();
             let typo = UniCase::new(typo);
-            rows.entry(typo)
-                .or_insert_with(BTreeSet::new)
-                .extend(i.map(|c| {
-                    let mut c = c.to_owned();
-                    c.make_ascii_lowercase();
-                    c
-                }));
```
epage: This change is incorrect.

This will cause the last implementation of a typo to "win" when the intention is the merge the entries

crates/typos-dict/tests/verify.rs
```diff
@@ -74,14 +69,16 @@ fn generate<W: std::io::Write>(file: &mut W, dict: &[u8]) {
...
+            let new_corrections: BTreeSet<_> = corrections
+                .into_iter()
+                .map(|correction| {
+                    word_variants
+                        .get(&*correction)
+                        .and_then(|variants| find_best_match(&typo, correction.as_str(), variants))
+                        .map(|s| s.to_owned())
+                        .unwrap_or(correction)
+                })
+                .collect();
```
epage, line 115: imo this is not a definitive enough of an improvement to make

Relying on combinators too much can make the code harder to read, debug, etc.

crates/typos-dict/tests/verify.rs
not-my-profile: Ah good catch, fixed.

crates/typos-dict/tests/verify.rs
not-my-profile: It's a preliminary refactor for ce3da1f38cc0a068ae433701067d0af320c9f347 of #765, which changes the BTreeSet to a Vec ... with the combinator approach duplicates can then simply be removed by adding a `.unique()` from Itertools, which I think is a bit more elegant than coding the hashset explicitly.

epage: I'll also add overuse of combinators make refactors harder.

Relying further on a bad change doesn't make it better.

(that use of `unique` is also incorrect, its in the wrong place.  Switching to an `IndexSet` is a more trivial change and would be correct)

not-my-profile: Fair enough ... I've now also dropped this change.

I was hesitant to add a new dependency on indexmap but I guess we already indirectly depend on it via the human-panic dependency of typos-cli.

epage: We also directly depend on `toml` which depends on `toml_edit` which gives us `indexmap`

crates/typos-dict/tests/verify.rs
```diff
@@ -9,38 +9,71 @@ type Dict = BTreeMap<UniCase<String>, BTreeSet<String>>;
...
+            (
+                typo.to_owned(),
+                iter.map(ToOwned::to_owned).collect::<Vec<_>>(),
+            )
         });
+    dict_from_iter(iter)
+}
 
-    let rows: Dict = rows
+fn dict_from_iter<S: Into<String>>(
```
epage: I think I'd prefer `parse` to return a `Vec` and for `process` to accept it.

not-my-profile: Good call, this also lets us easily test that duplicate entries are merged ... I made that change and added that test :)

PR #789, opened 2023-08-03
Mention hidden files in github-action.md

szepeviktor:
Warn users that the action will not check hidden files.
For example `.github/*`

### DIFF
--- a/docs/github-action.md
+++ b/docs/github-action.md
@@ -53,7 +53,7 @@ suggestions based on that local diff](https://github.com/getsentry/action-git-di
 
 | Name | Description | Required | Default |
 |------|-------------|----------|---------|
-| files| Files or patterns to check | false | If not defined, entire repository is checked |
+| files| Files or patterns to check | false | If not defined, entire repository is checked except hidden files |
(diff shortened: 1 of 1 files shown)


docs/github-action.md
```diff
@@ -53,7 +53,7 @@ suggestions based on that local diff](https://github.com/getsentry/action-git-di
 
 | Name | Description | Required | Default |
 |------|-------------|----------|---------|
-| files| Files or patterns to check | false | If not defined, entire repository is checked |
+| files| Files or patterns to check | false | If not defined, entire repository is checked except hidden files |
```
epage, line 56: Except I believe the config would be respected for any excludes.

Let's instead say "If not defined, the default set of files are checked"

szepeviktor: Done 🍏

szepeviktor: I just badly need git index-based checking instead of filesystem-based checking.

epage:
Thanks!

szepeviktor:
Glad to contribute.

PR #790, opened 2023-08-05
Add support for `pyproject.toml` files

### LINKED ISSUE #361: Support for pyproject.toml
Is it possible to use the `pyproject.toml` directly instead of creating a specific `typos.toml` file?

dosisod:
This PR adds support for parsing `pyproject.toml` config files. The convention for these files is to put any tooling related configuration into the `tool.NAME` section, so in this case, `tool.typos`. I have verified that the changes are pulled correctly even if the `tool.typos` section is not present.

Closes #361.

PS: I am new to Rust, so let me know if there are any improvements that can be made! I'm new to Serde as well, so I'm unfamiliar with what options do/don't need to be there (I just copied the traits on the `Config` struct).

### DIFF
--- a/crates/typos-cli/src/config.rs
+++ b/crates/typos-cli/src/config.rs
@@ -36,13 +36,14 @@ pub struct PyprojectTomlTool {
 
 impl Config {
     pub fn from_dir(cwd: &std::path::Path) -> Result<Option<Self>, anyhow::Error> {
-        let config = if let Some(path) = find_project_file(cwd, SUPPORTED_FILE_NAMES) {
-            log::debug!("Loading {}", path.display());
-            Self::from_file(&path)?
-        } else {
-            None
-        };
-        Ok(config)
+        for file in find_project_files(cwd, SUPPORTED_FILE_NAMES) {
+            log::debug!("Loading {}", file.display());
+            if let Some(config) = Self::from_file(&file)? {
+                return Ok(Some(config))
+            }
+        }
+
+        Ok(None)
     }
 
     pub fn from_file(path: &std::path::Path) -> Result<Option<Self>, anyhow::Error> {
@@ -55,7 +56,13 @@ impl Config {
         })?;
 
         if path.file_name().unwrap() == "pyproject.toml" {
-            return Ok(toml::from_str::<PyprojectTomlConfig>(&s)?.tool.typos);
+            return match toml::from_str::<PyprojectTomlConfig>(&s) {
+                Err(_) => {
+                    log::debug!("No `tool.typos` section found in `pyproject.toml`, skipping");
+                    Ok(None)
+                },
+                Ok(config) => Ok(config.tool.typos)
+            }
         }
 
         match Self::from_toml(&s) {
@@ -478,15 +485,16 @@ impl DictConfig {
     }
 }
 
-fn find_project_file(dir: &std::path::Path, names: &[&str]) -> Option<std::path::PathBuf> {
+fn find_project_files(dir: &std::path::Path, names: &[&str]) -> Vec<std::path::PathBuf> {
     let mut file_path = dir.join("placeholder");
+    let mut found_files = vec![];
     for name in names {
         file_path.set_file_name(name);
         if file_path.exists() {
-            return Some(file_path);
+            found_files.push(file_path.clone());
         }
     }
-    None
+    found_files
 }
 
 impl PartialEq for DictConfig {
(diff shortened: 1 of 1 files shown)


crates/typos-cli/src/config.rs
```diff
@@ -30,6 +45,11 @@ impl Config {
 
     pub fn from_file(path: &std::path::Path) -> Result<Self, anyhow::Error> {
         let s = std::fs::read_to_string(path)?;
+
+        if path.file_name().unwrap() == "pyproject.toml" {
+            return Ok(toml::from_str::<PyprojectTomlConfig>(&s)?.tool.typos);
+        }
+
         Self::from_toml(&s)
```
epage, line 53: Wrote up some thoughts in #361.  I originally thought we were going to need dicussion beyond this PR but I think we can handle it within this PR.

If `pyproject.toml` does not have a `tool.typos` table, we should keep looking for a config.

To do this, we'll need
- `pub typos: Config` to be `pub typos: Option<Config>`
- Either `from_file` to now return `Option<Self>` or for a new `from_file_` that returns `Option<Self>` with `from_file` doing a `unwrap_or_default`

dosisod: Sorry for the delay on this. I just pushed a commit that should address this.

One thing I ran into is that `find_project_file` is called before `from_file`, so it is hard to figure out what the next config file should be. I fixed this by making `find_project_file` return a list of files instead, that way `from_dir` can iterate over each `from_file` result and handle the potential `None` type.

Actually, I'm just now realizing I moved `pyproject.toml` to the front of `SUPPORTED_FILE_NAMES` for testing purposes then moved it to the back before pushing, so this looping logic might not be needed anymore. If we happen to rearrange `SUPPORTED_FILE_NAMES` though, or want to support more language-specific files in the future, then the for loop should probably stay. What are your thoughts?

crates/typos-cli/src/config.rs
```diff
@@ -19,26 +20,55 @@ pub struct Config {
...
+                    Ok(None)
+                },
+                Ok(config) => Ok(config.tool.typos)
+            }
+        }
+
+        match Self::from_toml(&s) {
+            Ok(x) => Ok(Some(x)),
+            Err(x) => Err(x)
+        }
```
epage, line 71: nit: imo these are two mutually exclusive business-logic paths (not just bookkeeping) and that should be reflected in the control flow with an `else`

crates/typos-cli/src/config.rs
```diff
@@ -19,26 +20,55 @@ pub struct Config {
...
                 format!("could not read config at `{}`", path.display()),
             )
         })?;
-        Self::from_toml(&s)
+
+        if path.file_name().unwrap() == "pyproject.toml" {
+            return match toml::from_str::<PyprojectTomlConfig>(&s) {
+                Err(_) => {
+                    log::debug!("No `tool.typos` section found in `pyproject.toml`, skipping");
+                    Ok(None)
```
epage, line 62: This will mask errors from deserializing the configuration

crates/typos-cli/src/config.rs
```diff
@@ -455,15 +485,16 @@ impl DictConfig {
...
 
-fn find_project_file(dir: &std::path::Path, names: &[&str]) -> Option<std::path::PathBuf> {
+fn find_project_files(dir: &std::path::Path, names: &[&str]) -> Vec<std::path::PathBuf> {
     let mut file_path = dir.join("placeholder");
+    let mut found_files = vec![];
     for name in names {
         file_path.set_file_name(name);
         if file_path.exists() {
-            return Some(file_path);
+            found_files.push(file_path.clone());
```
epage, line 494: With this new structure, there is no point in creating the mutable `file_path` but we can instead just `found_files.push(dir.join(name))`

crates/typos-cli/src/config.rs
```diff
@@ -19,26 +20,55 @@ pub struct Config {
...
+                Err(_) => {
+                    log::debug!("No `tool.typos` section found in `pyproject.toml`, skipping");
+                    Ok(None)
+                },
+                Ok(config) => Ok(config.tool.typos)
+            }
+        }
+
+        match Self::from_toml(&s) {
+            Ok(x) => Ok(Some(x)),
```
epage, line 69: nit: `Self::from_toml(&s).map(Some)`

crates/typos-cli/src/config.rs
epage: > If we happen to rearrange SUPPORTED_FILE_NAMES though, or want to support more language-specific files in the future, then the for loop should probably stay. What are your thoughts?

Yes, it seems appropriate for it to stay

crates/typos-cli/src/config.rs
epage, line 494: nit: we could make the return type `impl Iterator<Item=PathBuf>` if we make the function body
```rust
names.iter().map(|name| dir.join(name)).filter(|path| path.exists())
```

epage:
Thanks!

Borda:
This is great! Can we add an example of how this could be used?

PR #792, opened 2023-08-07
fix(dict): Remove only corrections if a space could be inserted as well

not-my-profile:
The typo dictionary words.csv previously contained a bunch of problematic entries such as:

    abouta,about
    algorithmi,algorithm
    attachen,attach
    shouldbe,should

Which resulted in wrong corrections if the following spaces (indicated by ␣) were accidentally missed:

    about␣a
    algorithm␣i developed
    attach␣en masse
    should␣be

Many of these entries were introduced by taking entries from the codespell-dict and removing corrections containing spaces (since typos currently doesn't support them), e.g the codespell dictionary contains:

    abouta->about a, about,
    shouldbe->should, should be,

This commit updates `tests/verify.rs` to automatically remove entries in the form of `{correction}{common_word},{correction}`, where `{common_word}` is one of the 1000 most frequent English words.

The top-1000-most-frequent-words.csv file was generated by running:

    curl https://norvig.com/ngrams/count_1w.txt \
      | head -n1024 \
      | awk '{print $1;}' \
      | grep -vE '^([^ia]|al|re)$' \
      > top-1000-most-frequent-words.csv


crates/typos-dict/assets/words.csv
```diff
@@ -24551,7 +24533,6 @@ extremaly,extremely
 extremaste,extremes
 extremeley,extremely
 extremelly,extremely
-extrememe,extreme
 extrememely,extremely
```
epage, line 24536: Let's revert this one because there are related corrections for this

crates/typos-dict/assets/top-1000-most-frequent-words.csv
```diff
@@ -0,0 +1,1000 @@
+the
```
epage, line 1: Let's leave this off for now because we'd need to workout cases like "extrememe"

crates/typos-dict/assets/words.csv
```diff
@@ -139,7 +139,6 @@ abotu,about
 abou,about,abound
 abount,about
 abourt,abort,about
-abouta,about
```
epage, line 142: Alternatives
- We report these as bad without suggesting a correction
  - We support this elsewhere though we might not support this here.  So could be easy, could be hard.
- We do a best-effort on joining the words
  - This would work in `abouta_foo` but not `abouta foo`
  - This would be a lot more work

Thoughts?

not-my-profile: Yes I did already think about reporting these as bad without suggesting a correction.
How about allowing corrections to contain spaces but if they do just not applying the corrections automatically?

So we could have e.g. `abouta,about,about a`, which would become:

```
error: `abouta` isn't a word, did you mean `about` or `about a`?
```

crates/typos-dict/assets/top-1000-most-frequent-words.csv
not-my-profile: I worked these cases out by adding the check:

```rs
                        if only_correction.ends_with(suffix) {
                            // We still want to correct e.g. "extrememe" to "extreme".
                            return true;
                        }
```

crates/typos-dict/assets/words.csv
epage: > How about allowing corrections to contain spaces but if they do just not applying the corrections automatically?

Thats much more invasive and I'd liek to hold off for now.  Something like that should be discussed in an issue first so we can plan out different aspects before focusing on the implementation

not-my-profile: But I think we can address this in a followup PR. Not suggesting corrections that are likely to be wrong is still better than missing a potential typo since false automatic corrections are much more undesirable than just false negatives.

not-my-profile: I just opened #795 for tracking that. Are you alright with removing these corrections for now?

epage: With a small change, we now support disallowed words in the dictionary.  We should at least do that within this PR rather than completely drop them

not-my-profile: Oh yeah thanks, that does make much sense and is a very elegant solution.

not-my-profile:
I have updated the PR to now also detect `{common_word}{correction},{correction}` and not just `{correction}{common_word},{correction}`.

crates/typos-dict/assets/words.csv
```diff
@@ -1,32 +1,32 @@
...
-aactually,actually
-aadd,add
-aadded,added
-aadding,adding
-aagain,again
-aaggregation,aggregation
-aand,and
-aanother,another
-aapply,apply
+aack
```
epage, line 5: The redundant a's is a separate thing and we should be correcting these

The challenge with blindly checking concatenated words is it doesn't filter out for when they don't make sense.

crates/typos-dict/assets/top-1000-most-frequent-words.csv
epage: I think I'd still prefer not be constrained by this very mechanical process.  It can provide insight but I don't trust it to automatically be applied

crates/typos-dict/assets/words.csv
epage: In general, I'd feel better if we just looked at what changed due to  the spaces and applied it to those.  We could then separate decide which of these changes might make sense.  As is, I'm seeing a lot that don't and don't want to take the time to decide that.

not-my-profile: Oh the logic I had did actually already detect these with

```rs
            if only_correction.starts_with(prefix) {
                return false;
            }
```

I just must have omitted resetting `words.csv` to run `SNAPSHOTS=overwrite cargo test verify` again.

not-my-profile: So the many cases have just become:

```
aequidistant
aequivalent
afor
amuch
anumber
ascripts
asudo
imakes
isimilar
itheir
itheirs
iwithout
```

which I think make sense to not correct automatically.

crates/typos-dict/assets/top-1000-most-frequent-words.csv
not-my-profile: The current process now seems to make exactly the changes we want to all our 63,200 entries, which does inspire some confidence in me. Besides the process is very easy to adapt so I think we can just do so when we figure out that it too eagerly filters out something.

epage: > The current process now seems to make exactly the changes we want to all our 63,200 entries

Except this isn't exactly what I want (see the other thread).  Arbitrarily combining words that don't make sense when combined leads to us losing corrections we would have otherwise.

crates/typos-dict/assets/words.csv
epage: `iwithout` and many of these don't make sense from a "word combining" perspective.  Things we can correct in master will become uncorrectable with this change.

PR #793, opened 2023-08-07
refactor(cli): Introduce file_type_specifics module

### LINKED ISSUE #759: Simplify contributing to default type-specific dictionaries
The type-specific dictionaries are currently hard-coded in a quite verbose manner, e.g:

https://github.com/crate-ci/typos/blob/84c8b30e065012c70187818aadba0e711ff5cafc/crates/typos-cli/src/config.rs#L210-L226

I think it would be nice to instead have a config file such as:

```toml
[rust.extend_identifiers]
flate2 = "flate2"
[rust.extend_words]
ser = "ser"
```

Just to make contributing such defaults more convenient. The verbose code could still be generated via a [build script](https://doc.rust-lang.org/cargo/reference/build-scripts.html). An alternative could be a declarative macro, however I think a dedicated .toml file would be nicer because it would enable other projects to easily use the data as well.

not-my-profile:
Resolves #759.

### DIFF
--- a/crates/typos-cli/src/config.rs
+++ b/crates/typos-cli/src/config.rs
@@ -2,6 +2,8 @@ use std::collections::HashMap;
 
 use kstring::KString;
 
+use crate::file_type_specifics;
+
 pub const SUPPORTED_FILE_NAMES: &[&str] = &["typos.toml", "_typos.toml", ".typos.toml"];
 
 #[derive(Debug, Clone, Default, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
@@ -155,9 +157,11 @@ pub struct TypeEngineConfig {
 
 impl TypeEngineConfig {
     pub fn from_defaults() -> Self {
-        let patterns = [
-            (
-                KString::from("lock"),
+        let mut patterns = HashMap::new();
+
+        for no_check_type in file_type_specifics::NO_CHECK_TYPES {
+            patterns.insert(
+                KString::from(*no_check_type),
                 GlobEngineConfig {
                     extend_glob: Vec::new(),
                     engine: EngineConfig {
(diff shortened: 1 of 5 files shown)


crates/typos-cli/src/file_type_specifics.rs
```diff
@@ -0,0 +1,76 @@
+//! This module specifies [`EngineConfig`] defaults for the file types defined in [`default_types`].
```
not-my-profile, line 1: I think it would make even more sense to move this module to the `typos-dict` crate ... however then we should also move the `default_types` module ... what do you think about that?

crates/typos-cli/src/config.rs
```diff
@@ -155,93 +157,49 @@ pub struct TypeEngineConfig {
...
-                GlobEngineConfig {
-                    extend_glob: Vec::new(),
-                    engine: EngineConfig {
-                        dict: Some(DictConfig {
-                            extend_identifiers: maplit::hashmap! {
-                                "windo".into() => "windo".into(),
-                            },
+        let patterns = file_type_specifics::NO_CHECK_TYPES
+            .iter()
+            .map(|no_check_type| {
```
epage, line 162: This is doing too much work with iterator combiantors which I find tends to lead to less readable, less debugable, and harder to refactor code.

crates/typos-cli/src/file_type_specifics.rs
```diff
@@ -0,0 +1,66 @@
...
+    ),
+];
+
+pub struct StaticDictConfig {
+    pub ignore_idents: &'static [&'static str],
+    pub ignore_words: &'static [&'static str],
+}
+
+#[cfg(test)]
+#[test]
```
epage, line 56: nit: technically, this is redundant

crates/typos-cli/src/file_type_specifics.rs
```diff
@@ -0,0 +1,66 @@
...
+fn test_type_specific_dicts() {
+    let types: Vec<_> = TYPE_SPECIFIC_DICTS.iter().map(|(typ, _)| typ).collect();
+    for types in types.as_slice().windows(2) {
```
epage, line 59: If you sort the list and render it into a newline separated string, you could get a nice view of the whole thing at once with `snapshot::assert_eq`

crates/typos-cli/src/file_type_specifics.rs
epage: For now, let's leave it here so we can get a better idea of usage.  For example, do we want per-format config besides dictionary information?

epage: (which the answer is already "yes" with `NO_CHECK_TYPES`)

epage: If/when we move some of this information, it wouldn't be in an array but in individual variables that we load up so nothing is tied to `default_types`

crates/typos-cli/src/file_type_specifics.rs
not-my-profile: Yeah ... but I don't really see the advantage of that. The current code is already simple and produces nice error messages.

epage: I disagree about the current code being simple

not-my-profile: Ok, point taken. I agree that with `.join("\n")` and `snapbox::assert_eq` the test code would become simpler. However the error messages would as I see it become worse. Compare:

```
thread 'file_type_specifics::test_type_specific_dicts' panicked at 'Eq failed: 
--- Expected
+++ Actual
   1    1 | py
        2 + rust
   2    3 | vim
   3      - rust
   4    4 | vimscript∅
```

with:

> thread 'file_type_specifics::test_type_specific_dicts' panicked at '"rust" should come before "vim" in TYPE_SPECIFIC_DICTS', crates/typos-cli/src/file_type_specifics.rs:65:9

Note how the current error message mentions `TYPE_SPECIFIC_DICTS`. Unless I am missing something [snapbox::Assert](https://docs.rs/snapbox/latest/snapbox/struct.Assert.html) does not appear to support customized error messages.

crates/typos-cli/src/file_type_specifics.rs
not-my-profile: Oh ... I didn't know that. Yeah that makes sense. I removed the `#[cfg(test)]`.

crates/typos-cli/src/config.rs
not-my-profile: Oh yeah, I should have remembered your distaste for iterator chains. I changed it to just mutate the HashMap.

crates/typos-cli/src/file_type_specifics.rs
not-my-profile: I don't see any reason why we shouldn't move `default_types` to the `typos-dict`. But yeah I'm fine with leaving this here for now.

crates/typos-cli/src/file_type_specifics.rs
epage: I think its good enough if `test_type_specific_dicts` is renamed to `test_type_specific_dicts_is_sorted`

crates/typos-cli/src/config.rs
epage: Iterator chains are fine, if they are simple.

crates/typos-cli/src/file_type_specifics.rs
not-my-profile: The test does however not only check if the slice is sorted but also that it does not contain duplicate entries.

My current code provides a different error message depending on whether the order is wrong or a duplicate entry exists, which I think is nice.

Yes we could have two test functions with largely the same code but at that point I'd argue that my current code is more elegant.

epage: In a lot of cases, I try to defer to the implementer as they are the ones who is immediately making a change.  For example, I had overlooked the de-duping in the name.

However, in the case of how this is handled, what I do not want to see
- Manual sorting checks
- Manual de-duping checks

And ideally, we don't show people a pigeon-hole view of the problem but that they have to fix one step at a time but we show them the whole set.  There are ways my suggestion can work between how the function is named and the naming of the variabled passed to the assert.  Alternatively, I'm fine with something like [pretty_assertions](https://crates.io/crates/pretty_assertions) being used.

So please change this!

not-my-profile: Thanks for your attention to detail and elaborating your reasoning. That does make a lot of sense. I split the test in two (which I now realize is better anyway) and updated it to use `snapbox::assert_eq` and `.join("\n")` like you initially suggested.

epage:
Thanks for your work on this!

PR #796, opened 2023-08-08
fix(dict): Allow `spawnve`

not-my-profile:
The first commit sorts `allowed.csv` since it's starting to get uncomfortable for humans.

### DIFF
--- a/crates/typos-dict/assets/allowed.csv
+++ b/crates/typos-dict/assets/allowed.csv
@@ -12,6 +12,7 @@ pervious,adjective
 perviously,adverb
 referer,http header field
 simulative,adjective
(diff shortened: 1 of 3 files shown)


crates/typos-dict/assets/allowed.csv
```diff
@@ -12,6 +12,7 @@ pervious,adjective
 perviously,adverb
 referer,http header field
 simulative,adjective
+spawnve,Microsoft-specific function (see https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/spawnve)
```
epage, line 15: I'm assuming, through bindings, this appears in enough programming languages that this is a generally allowed word, rather than a file-type specific one?

PR #797, opened 2023-08-08
refactor(cli): Make `file` module more DRY

not-my-profile:


epage:
Not a big fan of the coupling that most of these refactors introduce.  DRY is not a goal in of itself but a way to look at things to possibly improve code and needs to be balanced with other design principles.  In the future, it'd help if the focus was on the actual impact of the change rather than buzzwords like DRY.

I've cleaned up the part of this that I'm fine moving forward with in #798.

epage:
074a259  might work as constructors on `Typo`
- `extract_line` should move with it
- the constructor should have no knowledge of `AccumulateLineNum` but should instead accept `line_num`

Even then, I'm a bit hesitant.

epage:
> On a related note should we rename reporter::Typo to ReporterTypo? I find having two types with the same name to be a bit confusing.

Thats a weird one in the Rust community.  Some feel that module names are sufficient and applying prefixes is superfluous while others like to `use` everything and then prefixes are needed.  I'm not quite sure where I land.  I try to minimize my `use`s as much as possible as they are annoying to keep up to date with changes, a source of merge conflicts, and can easily obscure whats happening in the code by removing important context.

Issue #823, opened 2023-09-01
Non-critical corrections - September 2023 Edition


epage:
Please list any corrections you noticed `typos` doesn't fix and we'll roll these up into a single PR, making it lighter weight for adding new corrections.

m417z:
Wrong corrections:
```
intercection -> interception // the correct word was intersection
appliable -> applicable // appliable was the correct word
```

jrosdahl:
`failes -> fails, failed` (not just fails)

WalkerGriggs:
A few I picked up recently

```
concatincate -> concatenate
interleiive -> interleave
interleive -> interleave
transcdig -> transcoding
trasncoding -> transcoding
usses -> uses
```

szepeviktor:
```
recuse -> recurse
```

mo8it:
```
opionated -> opinionated
importantely -> importantly
stads -> stats, stands
```

not-my-profile:
```
impleiment -> implement
```

SOF3:
treaeted -> treated

ldirer:
`keeo -> keep`

phip1611:
`Hypevisor` -> `Hypervisor`

jplatte:
negligeable -> negligible

SOF3:
fedreated -> federated

cmdcolin:
potential wrong correction: typos currently changes

alinging ->alining

update to

alinging -> aligning

dupgit:
New typo for verbose:

verboce -> verbose

adamchalmers:
interleaed -> interleaved

WalkerGriggs:
```
rendtion -> rendition
```

jgillard:
```
extenally -> externally
intenally -> internally
```

epage:
From rust-lang/cargo#12714
```
unknowningly -> unknowingly
unecrypted -> unencrypted
registres -> registries
```

jrosdahl:
> ```
> registres -> registries
> ```

I guess that could be `registers` as well.

epage:
Updated it

jrosdahl:
```
assued -> assumed
```

cbachhuber:
preinted -> printed

Delgan:
```
usefull -> useful
```

[...]

PR #829, opened 2023-09-13
Adds 'verboce' as a typo for 'verbose'.

dupgit:
Hi,

I propose to add this typo ('verboce' for 'verbose') in the patch included.

Regards.


epage:
To double check, is there a reason this needs to go in now vs being rolled up in https://github.com/crate-ci/typos/issues/823 ?

dupgit:
> To double check, is there a reason this needs to go in now vs being rolled up in #823 ?

No there is no need for this patch to go in now. It can go into #823.  I'll report it there. Please forgot this commit.

Regards.

PR #834, opened 2023-09-26
adding docs about using `pyproject.toml`

Borda:
Adding note about using `pyproject.toml` which was added recently:
ref:
- https://github.com/crate-ci/typos/pull/790#issuecomment-1735033152
- https://github.com/crate-ci/typos/issues/361#issuecomment-1712008282


Borda:
@epage mind have look and approve CI run, pls :rabbit:

README.md
```diff
@@ -103,6 +103,22 @@ If you need some more flexibility, you can completely exclude some files from co
...
+extend-ignore-identifiers-re = ["AttributeID.*Supress.*"]
+
+[tool.typos.default.extend-identifiers]
+AttributeIDSupressMenu = "AttributeIDSupressMenu"
+
+[tool.typos.default.extend-words]
+teh = "teh"
+```
+
 ### Integrations
```
epage, line 122: FYI this was covered in #826.  If we want to expand it, we should probably do it in that area within the context of whats there

Borda: I may suggest to have it readme as it is easy to find and the other location was not discoverable =( so I found it based on issue and merged PR...

epage: The README is focused on being a quick start, delegating to this file for the details.

Borda: Yes but for some cases like us, the need to add a new config in the root could be a deal breaker so in such cases we cannot use it...

README.md
```diff
@@ -103,6 +103,22 @@ If you need some more flexibility, you can completely exclude some files from co
...
+```toml
+[tool.typos.default]
+extend-ignore-identifiers-re = ["AttributeID.*Supress.*"]
+
+[tool.typos.default.extend-identifiers]
+AttributeIDSupressMenu = "AttributeIDSupressMenu"
+
+[tool.typos.default.extend-words]
+teh = "teh"
+```
```
Borda, line 120:
```suggestion
In addition, you can set the same configuration with standard python's config file `pyproject.toml` by adding `[tool.typos.xyz]` where the `xyz` is the same as in default `_typos.toml`.
```
make it lighter

epage:
This is not aligning with the goal of this section of the documentation.  This is a getting started and isn't meant to cover every feature people might need to use.

Borda:
> This is not aligning with the goal of this section of the documentation. This is a getting started and isn't meant to cover every feature people might need to use.

do you call configuration a feature? :)

README.md
mecampbellsoup: also the docs added in #826 are crappy and insufficient. in general this project suffers from not having enough examples to get up the learning curve _immediately_

Issue #347, opened 2021-09-15
Excluding files unexpected behavior.


AIGeneratedUsername:
**Current settings:**

__typos.toml_
```toml
[files]
extend-exclude = ["*.json", "**/*.json"]
```

_.pre-commit-config.yaml_
```
-   repo: https://github.com/crate-ci/typos
    rev: typos-v0.7.4
    hooks:
      - id: typos
```

**Expected behavior:**
JSON files must be excluded 

**Expected behavior:**
All files are checked.

**Steps to reproduce**

` pre-commit run typos --all-files`

epage:
`typos` assumes the user knows best when passing in file paths, overriding the exclusions.  Unfortunately, this doesn't layer well with tools like pre-commit.

We need to provide a way to control this behavior.

amitlevy21:
I have the same issue, but only when running `typos-src` with `pre-commit`, running typos from the terminal works fine.
this is my config (_typos.toml at project root directory):
```toml
[files]
extend-exclude = ["src/assets/locales"]

```

locale files are ignored when running from the terminal, but not from `pre-commit run --all-files -v typos-src`

```yaml
  - repo: https://github.com/crate-ci/typos
    rev: v1.3.3
    hooks:
      - id: typos-src
        # args:
        #   - --exclude
        #   - src/assets/locales #uncommenting args works, but its not reading my config file

```

kwanhur:
> `typos` assumes the user knows best when passing in file paths, overriding the exclusions. Unfortunately, this doesn't layer well with tools like pre-commit.
> 
> We need to provide a way to control this behavior.

What about the hook attribute [exclude](https://pre-commit.com/#hooks-exclude) ? It can exclude files with specified pattern.

TomaszAIR:
Hi, we have a little bit different situation

typos.toml

```
$ cat typos.toml 
[files]
extend-exclude = ["Cargo.lock"]
```
.pre-commit-config.yaml

```
repos:
  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v4.1.0
    hooks:
      - id: check-added-large-files
      - id: check-merge-conflict
      - id: check-symlinks
      - id: detect-private-key
      - id: end-of-file-fixer
      - id: trailing-whitespace
  - repo: https://github.com/crate-ci/typos
    rev: v1.4.1
    hooks:
      - id: typos
```

Running couple of times `pre-commit run typos --all-files` gives different results. 2 passes and 1 fail

```
[...]

PR #837, opened 2023-09-30
Add "--force-exclude" option

Delgan:
Hi.

As discussed in #347, I'm proposing here a simple `--force-exclude`  option if that sounds good to you.

Regarding the implementation, I had to use [`Override.matched()`](https://docs.rs/ignore/0.4.20/ignore/overrides/struct.Override.html#method.matched) because the [`Walk`](https://docs.rs/ignore/0.4.20/ignore/struct.Walk.html) class unconditionally includes the initial path it receives. This means that any explicitly passed file we might want to exclude would not have a chance to be filtered.

I'm not very familiar with Rust so let me know if there are things you want me to improve.


crates/typos-cli/src/config.rs
```diff
@@ -98,6 +98,8 @@ impl Config {
 #[serde(rename_all = "kebab-case")]
 pub struct Walk {
     pub extend_exclude: Vec<String>,
+    // Exclude files even if passed explicitly.
+    pub force_exclude: Option<bool>,
```
epage, line 102: To me, this feature seems more situational rather than being a fixed state, so I think it should be cli only and not in config

crates/typos-cli/tests/cmd/force-exclude.toml
```diff
@@ -0,0 +1,6 @@
+bin.name = "typos"
+args = "file.ignore --force-exclude"
```
epage, line 2: Can you add this test, without the flag, in its own commit, ensure it passes, then implement this feature and update the test?

This helps test the test and show to reviewers that it all works as expected.

crates/typos-cli/src/bin/typos-cli/main.rs
```diff
@@ -216,6 +216,12 @@ fn run_checks(args: &args::Args) -> proc_exit::ExitResult {
             let overrides = overrides
                 .build()
                 .with_code(proc_exit::sysexits::CONFIG_ERR)?;
+            if walk_policy.force_exclude() && path.is_file() {
```
epage, line 219: Why the file check?

crates/typos-cli/src/bin/typos-cli/main.rs
```diff
@@ -216,6 +216,12 @@ fn run_checks(args: &args::Args) -> proc_exit::ExitResult {
...
+                match overrides.matched(path, false) {
```
epage, line 220: Something I need to double check is that some of these APIs expect you to call them as you recurse into a directory and return confusing results if you only call them on the final path.

epage (commented):
Thanks for moving this forward!

crates/typos-cli/src/bin/typos-cli/main.rs
Delgan: Good question... I think I over-focused on my specific use case, sorry.

crates/typos-cli/tests/cmd/force-exclude.toml
Delgan: Sure, I pushed a commit which removes the `--force-exclude` from the tests and should fail the CI.
(Although I forgot to revert the implementation as you suggested, but I'm not sure it's helpful as I assume the commits will be squashed before merge anyway).

crates/typos-cli/src/bin/typos-cli/main.rs
Delgan: Hum... I think you're correct. If I have `foo/bar/baz.file` directory structure and `extend-exclude = ["bar/*.file"]` configuration, using `typos foo` will ignore the file properly but `typos foo/bar/baz.file --force-exclude` won't (using `extend-exclude = ["foo/bar/*.file"]` or `extend-exclude = ["baz.file"]` works, though). :confused: 

Not sure how to fix it. Maybe we should propose a feature to `ignore` crate so that it optionally checks root entry as well: [`walk.rs#L936-L938`](https://github.com/BurntSushi/ripgrep/blob/f608d4d9b3ab210b7e6964ca7d1d7dc9c077329e/crates/ignore/src/walk.rs#L936-L938)?
I don't know how likely is it to be accepted, knowing that `Walk` is supposed to be used with directory as argument (from the docs at least), and we're using it with file path as well.

crates/typos-cli/tests/cmd/force-exclude.toml
epage: > l as I assume the commits will be squashed before merge anyway


imo squashing is an anti-pattern.  You either lose valuable information, locked away in a proprietary hosting service or you force people into making non-atomic PRs (refactors that don't make sense without other changes, etc).

Please swap this to the two commit approach *with tests passing in each commit*

crates/typos-cli/src/bin/typos-cli/main.rs
epage: You likely need to do `path.ancestors()`.rev()`, calling `matched` on each

Delgan: Apologies, I think I just did something wrong during my test. The file is **not** excluded in any case, which is the expected behavior.

crates/typos-cli/tests/cmd/force-exclude.toml
Delgan: > Please swap this to the two commit approach _with tests passing in each commit_

Sorry, I misinterpreted your instructions. It's done now.

> imo squashing is an anti-pattern. You either lose valuable information, locked away in a proprietary hosting service or you force people into making non-atomic PRs (refactors that don't make sense without other changes, etc).

To me, squashing serves to remove the commits that are just noise and do not belong to the `main` branch, in particular the ones occurring during the review process (such commits would not exist if the PR was _perfectly_ implemented in the first place).
The alternative is to rewrite the history in response to a comment, but then the whole commit has to be reviewed again instead of just the last change, which is a tedious. The "Compare" feature of Github kind of solves this problem, though.

Anyway, I updated the PR according to your preferences :+1:

crates/typos-cli/src/bin/typos-cli/args.rs
```diff
@@ -57,6 +57,10 @@ pub(crate) struct Args {
     #[command(flatten, next_help_heading = "Config")]
     pub(crate) config: ConfigArgs,
 
+    /// Respect excluded files even for paths passed explicitly.
+    #[arg(long, group = "mode", help_heading = "Mode")]
```
epage, line 61: Why is this in the `mode` group and under the `Mode` heading?

crates/typos-cli/tests/cmd/force-exclude.toml
epage: Thanks!

crates/typos-cli/src/bin/typos-cli/args.rs
Delgan: Please indicate the `group` and `help_heading` you prefer.

epage: It doesn't need to be in any group.  In fact, I expect the use of the group breaks the pre-commit hook because this declared `--write-changes` and `--force-exclude` as mutually exclusive.

As for a help_heading, let's leave this off for now, leaving it to `Options`

epage:
Sorry about the spelling failure in CI; I had a bad merge which introduced this workflow when it shouldn't be here

crates/typos-cli/src/bin/typos-cli/args.rs
Delgan: Thanks for the explanations. I removed the `group` and `help_heading` without moving the `arg` position, which means `--force-exclude` appears at the end of `Config` group. Let me know if you prefer otherwise (it would appear below the `-j` option if I move it up).

Delgan:
> Sorry about the spelling failure in CI; I had a bad merge which introduced this workflow when it shouldn't be here. Fixed in master

No problem. I rebased the PR. :+1:

crates/typos-cli/src/bin/typos-cli/args.rs
epage: Ah, I hadn't realized some config comes before this.  Please go with `help_heading = None` then

Delgan: I updated the PR. Is the output what you expected?

epage: Seems good enough

epage:
Thanks!

PR #844, opened 2023-10-02
feat(dict): September updates

### LINKED ISSUE #823: Non-critical corrections - September 2023 Edition

epage:
- The correction to `usefull` and `becuase` has existed from the initial implementation, as far as I could tell.
- `busses` is a valid variation of `buses`
- Had to put in a hack so that `aligning` would be preferred over `alining` because of how the preference algorithm works

Fixes #823

### DIFF
--- a/crates/typos-dict/assets/allowed.csv
+++ b/crates/typos-dict/assets/allowed.csv
@@ -17,3 +17,4 @@ unuseful,adjective
 contiguities,plural of contiguity
 bellow,valid word and not just a typo of `below`
 revered,valid word and not just a typoe of `reversed`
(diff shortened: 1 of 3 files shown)


szepeviktor:
With the current pace of adding words when do you expect Typos to get really slow?

epage:
> At the current rate of adding words, when do you expect Typos to get really slow?

Considering the dictionary is a trie, the slope of growth should be relatively shallow.

szepeviktor:
This release really works!
https://github.com/percona/percona-toolkit/pull/691/files

PR #849, opened 2023-10-09
fix(docs): typo in Config Fields

szepeviktor:
An "s" was missing.


epage:
Thanks for pointing this out but the right fix was 2b7af0e

Issue #386, opened 2021-12-03
Spell check commit messages via GitHub Action


kuhar:
It would be great if the action could also report typos in the commit message.

Currently, this can be worked around by saving the latest commit message to a file that is not a part of the checkout out repo.

epage:
I had originally considered building `typos` into [committed](https://github.com/crate-ci/committed) but checking the commit message in an action is a relatively low effort middle ground.

I'd either want this to be a separate action or have flags to turn on / off each.  Personally, I'd want a separate step or job for each so the feedback is more precise (one is more important than the other to me).

Similarly, we could add a pre-commit hook for this.

PR #856, opened 2023-10-17
fix(dict): Add Julia-specific dictionary

epage:
Inspired by JuliaLang/julia#51704


crates/typos-cli/src/default_types.rs
```diff
@@ -112,7 +112,6 @@ pub const DEFAULT_TYPES: &[(&str, &[&str])] = &[
     ("js", &["*.js", "*.jsx", "*.vue"]),
     ("json", &["*.json"]),
     ("jsonl", &["*.jsonl"]),
-    ("julia", &["*.jl"]),
```
epage, line 115: @LilithHafner would it be more idiomatic for the file type to be "julia" or "jl"?  Since we already refer to most files by their extension, rather than their full name, I went with `jl`.

Note this list was originally taken from ripgrep.  No idea why burntsushi made the decisions they did.

crates/typos-cli/src/default_types.rs
```diff
@@ -145,7 +144,7 @@ pub const DEFAULT_TYPES: &[(&str, &[&str])] = &[
         "OFL-*[0-9]*",
     ]),
     ("lilypond", &["*.ly", "*.ily"]),
-    ("lisp", &["*.el", "*.jl", "*.lisp", "*.lsp", "*.sc", "*.scm"]),
+    ("lisp", &["*.el", "*.lisp", "*.lsp", "*.sc", "*.scm"]),
```
epage, line 147: Explanatory note: this file was taken from ripgrep but we operate very differently.  With ripgrep, you do a run with a specific file type and it then directly loads the extensions,  We instead find the extensions and need to map that back to names to know what dictionaries to load.  We currently only support one mapping being used and have some bugs,m like this one, where we are allowing multiple to be specified when we can't handle it.

LilithHafner:
Does this mean any usage of "egals" in any file ending in `.jl` will be considered a word by default?

epage:
Yes

PallHaraldsson:
I find it kind of ironic that there's a typo "Alllow" in the commit message... :)

I'm not just being a dick pointing out, but I just heard of this project, and am thinking maybe it can, or could help with spellcheck commit messages too?

epage:
https://github.com/crate-ci/typos/issues/386 is at least one issue around commit messages.

PR #862, opened 2023-10-24
feat: Add commit-msg pre-commit hooks

sebasnabas:
Implements #386.

Important note: The hook in the `commit-msg` stage silently writes changes to the commit message.


epage:
Let's discuss possible solutions before moving on to PRs.

In this case, the request is general support while this is only helpful if you are using the pre-commit hook.

PR #864, opened 2023-10-27
fix(cli): Improve color formatting

### LINKED ISSUE #863: Readable colors
Hello! codespell has nicer colors. Yellow file names and red/green corrections.

image

How could I improve the colors of typos?

epage:
- Better matched rustc color choices (bright colors, rather than normal)
  - I intentionally kept to these colors, rather than the suggested yellow file path, because its more "positive" (yellow is more associated with negative things like warnings)
- Better matched rustc in what gets colored
- Went ahead and deviated from rustc in coloring the parts within the message
  - This works especially well for `--format brief` but I think it works well enough in `--format long` to justify the deviation

## `--format long`

Before

Screenshot from 2023-10-27 09-32-01

After

Screenshot from 2023-10-27 09-32-26

## `--format brief`

Before

Screenshot from 2023-10-27 09-32-42

After

Screenshot from 2023-10-27 09-32-55

Fixes #863


szepeviktor:
Thank you for the light blue!
I will be able to read that on black bg.

szepeviktor (approved):

PR #791, opened 2023-08-05
Minify size of release binaries

dosisod:
See https://github.com/johnthagen/min-sized-rust

Here are some stats on the different flags I used and how much each step reduced the binary size:

|           Method            |        Size        | Diff (%) |
|-----------------------------|--------------------|---------:|
| Master                      | 27633280 (27.6MB)  |      N/A |
| `strip`                     | 19544600 (19.5MB)  |   -29.3% |
| `strip` + `lto`             | 19028504 (19.0MB)  |   -31.1% |
| `strip` + `lto` + `codegen` | 18971160 (19.97MB) |   -31.3% |

Closes #738.

Issue #738, opened 2023-06-08
Please strip release binary


szepeviktor:
```
$ file typos
typos: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=8f7bf8a54b62d8ebc982815e30076433d398c44d, not stripped
```

The `typos` binary gets smaller when stripping.

epage:
#791 attempted to do what this issue directly asked.

I have not looked into stripping before but my concerns are
1. Quality of panic messages is not diminished
2. Users wanting to debug the binary can

Unless there is a reason we shouldn't do (2), that means we need to not just strip but we also need to make sure our release process includes the separate debug symbols in the payload.  At that point, I'm not as sure as to the benefit of doing this.

szepeviktor:
I am from an old world 👴🏻 
production was production, development was development.
For example Debian releases stripped binaries for production and debug symbol for debugging.

> [!INFO]
> Currently my `update-typos.sh` strips the binary

szepeviktor:
If you have _unqualified_ users please close this issue.
I know I live in the golden cage 🐦

dpc:
```
[profile.release]
debug = "line-tables-only"
```

+ 

```
[target.x86_64-unknown-linux-gnu]
rustflags = ["-C", "link-arg=-Wl,--compress-debug-sections=zlib"]
```

is a combo that became available in some recent Rust release I've been using nowadays that makes the binary smaller, yet preserves the (most essential?) symbols which makes stacktraces in errors and panics informative.

>  make sure our release process includes the separate debug symbols in the payload.

I'd be interested in how to get it done.

szepeviktor:
Thank you.

PR #880, opened 2023-12-05
bin: shrink binary size from 26MiB to 18MiB

phip1611:
The `typos` binary is relatively large. Currently, roughly `26MiB`. We can shrink that down to `18MiB`.

- status quo: `27807000` bytes / `26.52` MiB
- lto: `24852744` bytes / `23.70` MiB
- stripped: `19474952` bytes / `18.57` MiB
- stripped + lto: `18995720` bytes / `18.12` MiB

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,3 +27,5 @@ panic = "abort"
 [profile.release]
 panic = "abort"
 codegen-units = 1
+lto = true
+strip = true


epage:
Like with #791, we need to finish up discussin on #738 first.

phip1611:
Ohh, I didn't see there was an existing PR about that. Sorry.

PR #882, opened 2023-12-08
fix(config): Propagate validation errors

### LINKED ISSUE #881: Bug: invalid regex in `extend-ignore-identifiers-re` silently stops config file parsing
I found an error in the typos utility, I think. It seems that as soon as one entry in `extend-ignore-identifiers-re` is invalid

```toml
[default]
extend-ignore-identifiers-re = [
  # This entry causes harm; it should be "foo\\(" but no error is reported
  "foo(",
]
```

the parsing **silently** fails and causes that all other configurations, such as `[default.extend-identifiers]` and `[default.extend-words]` are **silently ignored**. That's very unfortunate.

One can verify this following these steps:

- create a `.typos.toml` with the following content:
  ```toml 
  [default]
  extend-ignore-identifiers-re = [
    # This entry causes harm; it should be "foo\\(" but no error is reported
    "foo("
  ]

  [default.extend-identifiers]
  bar = "foo"

  [default.extend-words]
  bar = "foo"
  ```
- run `typos --config .typos.toml --dump-config - | grep 'foo'`
- **error** you see no results
- remove the `foo(` in the config
- run `typos --config .typos.toml --dump-config - | grep 'foo'`
- it works

**Expected Behaviour**:

Typos should tell which regex is invalid and why it is invalid, instead of silently dropping that error. Furthermore, it should continue parsing the config file.

epage:
When `flatten`ing an `Option`, all errors are silenced, making the value
`None`, not just "this field doesn't exist".
The easiest way around this was to not use `Option`.

Not too confident in all of the changes but tests pass?

Fixes #881


phip1611:
Wonderful! Thanks for the quick fix :)

PR #886, opened 2023-12-27
fix(config): Apply extend-ignore-re to file names 

### LINKED ISSUE #885: Cannot exclude false positive in file names
```
error: `olt` should be `old`
  --> ./include/olt-manager.php:1
```

Please help me where to exclude "olt-manager".
It is bad practice to exclude "olt" in `extend-words`.
I'd like to exclude only the false positive in the file name, not its contents.

epage:
Fixes #885


szepeviktor (approved):
Thank you.

Issue #594, opened 2022-10-15
Support Sarif reporting format


halkeye:
[OASIS Static Analysis Results Interchange Format](https://www.oasis-open.org/committees/sarif/) is a newish standardization format for analysis tools.

It would be great if typos can support outputting that format.

Looks like there is already libraries to [generate the format](https://github.com/psastras/sarif-rs/blob/main/serde-sarif/src/converters/clang_tidy.rs) so it shouldn't be a hard lift.

halkeye:
I can make the attempt if desired, but I learned with the github actions formatter it might not be desirable.

epage:
Following the initial link, I only saw references to it being a draft and not finalized.  Though [other documents](https://sarifweb.azurewebsites.net/) I later found refer to it as approved with no draft mention.

The [docs for the Rust API](https://docs.rs/serde-sarif/latest/serde_sarif/) seem to caution use of the lib itself.

Also, any idea on how adoption of this has been so far?

halkeye:
> Also, any idea on how adoption of this has been so far?

I learned about it playing with codeql (eslint template) on github actions. They recommend https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/uploading-a-sarif-file-to-github

So it feels like microsoft and github are both pushing for it.

epage:
At times, it felt like I was reading about a security feature and at times it felt more broad.

The alert tracking sounds nice, like it might offer some of the static analysis benefits of a tool I managed at a prior job that allowed adding new static analysis without being buried under the weight of the backlog.

Overall, I would be in favor of this depending on the level of maturity of library support for it.  Depending on how this evolves, we'd need to be prepared for how we expose versioning.  Would we just do `serif`, `serif-2`, `serif-2.1` or something else?

andersk:
Since this is the typos project, I feel compelled to note: it’s SARIF, not SERIF. 😛

epage:
When coming up with a name for the project, I was tempted to make the name include a typo but figured that'd be too aggravating, either for people typing the command name (and spelling it correctly) or when running the command.

Zxilly:
I was wondering if anyone is working on this issue? I would like to add this support, but I am not a professional rust developer, so the code quality may be poor.

PR #937, opened 2024-02-14
feat: add Code Climate output format

afranke:
:wave: 

I’m adding the [Code Climate](https://github.com/codeclimate/platform/blob/master/spec/analyzers/SPEC.md#data-types) output format, that can then be used by other tools, such as [GitLab](https://docs.gitlab.com/ee/ci/yaml/artifacts_reports.html#artifactsreportscodequality), to display reports without having to dig in logs.

The work is not over yet, but I got to a point where I have something to show. I will improve on what I have, and welcome any feedback you may have for me.

One thing I’m not sure how to do yet is how to get a single JSON array as the output.

### DIFF
--- a/crates/typos-cli/src/bin/typos-cli/args.rs
+++ b/crates/typos-cli/src/bin/typos-cli/args.rs
@@ -10,6 +10,7 @@ pub enum Format {
     #[default]
     Long,
     Json,
+    CodeClimate,
 }
 
 impl Format {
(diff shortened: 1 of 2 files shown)


epage:
FYI for major changes, its generally best to start out with an Issue so we can make sure we are on the same page on whether this is a good fit and what considerations need to be taken into account.

For example, we have #594 for Sarif

epage:
Without the discussion on what we should be doing, I'm going to close this.

PR #973, opened 2024-04-02
fix(dict): Don't correct GUID

kachick:
Listed in #955, but GUID is used for abbreviation of Globally Unique Identifier, especially in Windows

Examples
---

* https://github.com/microsoft/terminal/blob/9f3dbab7bf442a1d7702bef0058fa2f0ff5736cd/src/types/utils.cpp#L48-L50

PR #965, opened 2024-04-02
fix(dict): Don't correct IME

kachick:
Listed in https://github.com/crate-ci/typos/issues/955, but `IME` is used as abbreviation of `Input Method Editor`

Examples
---

* https://github.com/microsoft/winget-pkgs/blob/bfe98ffdfa92825c9fc6ecb0975699b518749e97/manifests/g/Google/JapaneseIME/2.29.5370.0/Google.JapaneseIME.installer.yaml#L4
* https://github.com/grammarly/contenteditable/blob/b516071edd25a1df88c3441f0d71fe8f0c5f0cf4/README.md?plain=1#L30

PR #975, opened 2024-04-02
fix(dict): Avoid correcting common terms

### LINKED ISSUE #963: ws may be an abbreviation for websocket
Hello, I found that after the upgrade, typos will consider "ws" as a spelling error. However, in some scenarios, "ws" may be an abbreviation for WebSocket.

### LINKED ISSUE #970: Various false positives since 1.20
- `IIS` is corrected to `IS` (IIS is the webserver from Microsoft)
- `iot` is corrected to `it` (iot is short for internet of things)

Also, I found a few aggressive corrections. These corrections are probably fine, but I will list them nonetheless:
- `ang` is corrected to `and` (I used it as short for `angle`, but I probably should not short it.)
- `dur` is corrected to `due` (I used it as short for duration).

### LINKED ISSUE #974: `lst` could also mean "list"
`lst` is a common abbreviation for `list` that tends to be a keyword in many languages, it is incorrect to change it to "last" as done in #955.

epage:
Supersedes #965
Supersedes #973

Fixes #963
Fixes #974
Fixes #970

### DIFF
--- a/crates/typos-dict/assets/allowed.csv
+++ b/crates/typos-dict/assets/allowed.csv
@@ -24,3 +24,12 @@ eof,end-of-file in programming
 eol,end-of-line in programming
 og,OpenGraph which is used for previews of websites on social media
 jst,abbreviation of Japan Standard Time in tz database
+guid,globally-unique-identifier in programming
+ime,abbreviation of Input Method Editor
+ws,common abbreviation for websocket and workspace
+iot,acronym for Internet of Things
+iis,IIS is the webserver from Microsoft
+ro,acronym for read-only
+dur,abbreviation for duration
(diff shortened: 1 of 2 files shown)


crates/typos-dict/assets/allowed.csv
```diff
@@ -24,3 +24,12 @@ eof,end-of-file in programming
 eol,end-of-line in programming
 og,OpenGraph which is used for previews of websites on social media
 jst,abbreviation of Japan Standard Time in tz database
+guid,globally-unique-identifier in programming
+ime,abbreviation of Input Method Editor
```
svix-jplatte, line 28: Or "in my experience" (for capitalized IME) ^^

PR #983, opened 2024-04-05
Don't correct `HD`

### LINKED ISSUE #987: False positive caused by `HD`
## `HD` should be `HAD`

Example: https://github.com/DeadNews/shell-scripts/pull/25/commits/ea9d63cb3757c1d0318cefb414e6d1731dbe6858

`HD` is a common abbreviation.

https://en.wikipedia.org/wiki/HD#Technology
https://grep.app/search?q=hd&words=true

<details>

## `thr` should be `the`

Example: https://github.com/DeadNews/dnfunc/pull/147/commits/e0a7db6ba62f0a6ce95c090473fc1b9bf48b8d6c

`thr` is a common abbreviation for `threshold`.

https://grep.app/search?q=thr&words=true

</details>

----

I wonder if it is even worth fixing 2-3 letter sequences? 
The chance of false positives is very high. 
Is it possible to ignore all 2-3 letter sequences via regex in the settings?

alatiera:
Common abbreviation for High Definition

Fixes #987

### DIFF
--- a/crates/typos-dict/assets/words.csv
+++ b/crates/typos-dict/assets/words.csv
@@ -28613,7 +28613,6 @@ havve,have
 hax,hex
 haynus,heinous
 hazzle,hassle
-hd,had
 hda,had
 hdinsight,hindsight
 heacy,heavy
--- a/crates/typos-dict/assets/allowed.csv
+++ b/crates/typos-dict/assets/allowed.csv
@@ -1,6 +1,7 @@
 nilable,used in ruby community
 thead,html tag
 hardlinked,filesystem term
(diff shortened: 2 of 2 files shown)


crates/typos-dict/assets/words.csv
```diff
@@ -28613,7 +28613,6 @@ havve,have
 hax,hex
 haynus,heinous
 hazzle,hassle
-hd,had
```
epage, line 28616: Please add it to `allowed.csv` with an explanation.  This will prevent it from being added back in the future.

PR #985, opened 2024-04-08
Add more python file extensions

augustelalande:
Add more python file extensions. Namely `.ipynb` for jupyter notebook files, and `.pyi` for stub files.


augustelalande:
Just noticed jupyter notebook is a seperate entry. I also didn't realise jupyter supports more languages than just python (although I think python is still the most common). Anyway it would be nice so support the python ignores in jupyter, although I'm not sure how you guys want to structure that.

epage:
Let's split jupyter out to an issue to discuss how that should be handled and then we can move forward with `pyi` immediately

augustelalande:
Sounds like a plan

PR #1023, opened 2024-06-02
fix typos (the literal kind)

RainRat:


crates/varcon/assets/README
```diff
@@ -220,7 +220,7 @@ created by Garst R. Reese <reese@isn.net>:
   Canadians use ize ization, izing izable like Americans, and gram instead
   of gramme. The one exception I found was practise. It does not go to
   practize.  Otherwise they use British spelling. So, what I am currently
-  checking books with is a an edited version of British, where I have
+  checking books with is an edited version of British, where I have
```
epage, line 223: Please revert this as its a vendored file

RainRat: reverted

PR #1039, opened 2024-06-15
Add GitHub Actions with suggestions using reviewdog integration

haya14busa:
I created a yet another GitHub Actions which runs typos with reviewdog.
https://github.com/reviewdog/action-typos

It also support posting comments with GitHub Code Suggestions feature.

image

I'm not sure it's appropriate to add different actions in the documentation,
but I just wanted to share it as it's handy if suggestion feature is supported.

Please feel free to close it.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ Dual-licensed under [MIT](LICENSE-MIT) or [Apache 2.0](LICENSE-APACHE)
   - [False Positives](#false-positives)
   - [Integrations](#integrations)
     - [GitHub Action](docs/github-action.md)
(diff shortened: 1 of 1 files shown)


epage:
What would be the motivation for running (or when would we tell someone to run this) over the first-party action?

haya14busa:
I just wanted to share the created action because it also supports suggestion feature, which I feel very useful and I think other users also probably want this feature. 

If you're willing to manage the created action as a first party action, then it's also fine with me. 

As I commented above, please feel free to close it if you think otherwise.

PR #1040, opened 2024-06-17
add support for mac and arm systems to the github action

dyc3:
I've tested this on an M1 mac to make sure it works.

### DIFF
--- a/.github/workflows/test-action.yml
+++ b/.github/workflows/test-action.yml
@@ -4,7 +4,10 @@ on: [pull_request]
 jobs:
   shallow:
     name: Spell Check with Typos
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
     steps:
     - name: Checkout Actions Repository
       uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


epage:
Could you elaborate on motivation for this change?

dyc3:
Sure. I was setting up some CI for work, and we blew through the included CI minutes we had. I figured I'd move all of our CI to run on a self hosted runner, running on a mac mini. This was the only action that blew up when trying to run everything on that mac self hosted runner.

I believe this change would also make it possible for raspberry pis to run this action now, but I haven't tested it.

epage:
... your only self-hosted runner is a Mac?  That is the opposite of what I would expect since they have historically been more expensive runners that are harder to provision.  Like github, I've tended to see those kept to a minimum.  I'm assuming this is a shop focused specifically on Apple development??

epage:
We should update https://github.com/crate-ci/typos/blob/master/.github/workflows/test-action.yml to cover at least one of the newly supported arm platforms so we don't regress this

dyc3:
Yeah it's weird. Small startup and we need to do ios builds. I'll update that action.

dyc3:
I've updated the action. The `macos-latest` runner uses M1. See: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

.github/workflows/test-action.yml
```diff
@@ -4,7 +4,10 @@ on: [pull_request]
 jobs:
   shallow:
     name: Spell Check with Typos
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
```
epage, line 9: `macos-latest` is still x64, right?  To ensure we also cover arm, I think we'd need `macos-14`

epage:
Oh, hadn't kept track that `macos-latest` now points to `macos-14`!

PR #1047, opened 2024-06-28
feat: implement sarif format reporter

Zxilly:
Closes: #594 

I asked earlier if anyone had already worked on this feature and got no reply https://github.com/crate-ci/typos/issues/594#issuecomment-2179277793 , so I added this implementation myself.

But as I declared, I'm not a professional rust developer, so this patch obviously contains a lot of problems caused by a lack of understanding of rust, such as `.unwrap` abuse, since I'm not sure how to properly convert `Error` to `io::Error`.

I enabled `Allow edits by maintainers`, so if you have any suggestions for syntax changes, please push them directly to the corresponding repository branch, thanks!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,16 @@
 version = 3
 
 [[package]]
+name = "Inflector"
+version = "0.11.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
+dependencies = [
+ "lazy_static",
+ "regex",
+]
+
+[[package]]
 name = "addr2line"
 version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -123,7 +133,7 @@ checksum = "edf3ee19dbc0a46d740f6f0926bde8c50f02bdbc7b536842da28f6ac56513a8b"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.90",
 ]
 
 [[package]]
@@ -208,7 +218,7 @@ dependencies = [
  "anstream",
  "anstyle",
  "clap_lex",
- "strsim 0.11.0",
+ "strsim",
  "terminal_size",
 ]
 
@@ -221,7 +231,7 @@ dependencies = [
  "heck",
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.90",
 ]
 
 [[package]]
@@ -338,9 +348,9 @@ dependencies = [
 
 [[package]]
 name = "darling"
-version = "0.20.3"
+version = "0.20.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0209d94da627ab5605dcccf08bb18afa5009cfbef48d8a8b7d7bdbc79be25c5e"
+checksum = "6f63b86c8a8826a49b8c21f08a2d07338eec8d900540f8630dc76284be802989"
 dependencies = [
  "darling_core",
  "darling_macro",
@@ -348,27 +358,58 @@ dependencies = [
 
 [[package]]
 name = "darling_core"
-version = "0.20.3"
+version = "0.20.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "177e3443818124b357d8e76f53be906d60937f0d3a90773a664fa63fa253e621"
+checksum = "95133861a8032aaea082871032f5815eb9e98cef03fa916ab4500513994df9e5"
 dependencies = [
  "fnv",
  "ident_case",
  "proc-macro2",
  "quote",
- "strsim 0.10.0",
- "syn",
+ "strsim",
+ "syn 2.0.90",
 ]
 
 [[package]]
 name = "darling_macro"
-version = "0.20.3"
+version = "0.20.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "836a9bbc7ad63342d6d6e7b815ccab164bc77a2d95d84bc3117a8c0d5c98e2d5"
+checksum = "d336a2a514f6ccccaa3e09b02d41d35330c07ddf03a62165fcec10bb561c7806"
 dependencies = [
  "darling_core",
  "quote",
- "syn",
+ "syn 2.0.90",
+]
+
+[[package]]
+name = "derive_builder"
+version = "0.20.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "507dfb09ea8b7fa618fcf76e953f4f5e192547945816d5358edffe39f6f94947"
+dependencies = [
+ "derive_builder_macro",
+]
+
+[[package]]
+name = "derive_builder_core"
+version = "0.20.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2d5bcf7b024d6835cfb3d473887cd966994907effbe9227e8c8219824d06c4e8"
+dependencies = [
+ "darling",
+ "proc-macro2",
+ "quote",
+ "syn 2.0.90",
+]
+
+[[package]]
+name = "derive_builder_macro"
+version = "0.20.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ab63b0e2bf4d5928aff72e83a7dace85d7bba5fe12dcc3c5a572d78caffd3f3c"
+dependencies = [
+ "derive_builder_core",
+ "syn 2.0.90",
 ]
 
 [[package]]
@@ -388,7 +429,7 @@ checksum = "cb7330aeadfbe296029522e6c40f315320aba36fc43a5b3632f3795348f3bd22"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.90",
  "unicode-xid",
 ]
 
@@ -401,7 +442,7 @@ dependencies = [
  "darling",
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.90",
 ]
 
 [[package]]
@@ -447,7 +488,7 @@ checksum = "7bdb5411188f7f878a17964798c1264b6b0a9f915bd39b20bf99193c923e1b4e"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.90",
 ]
 
 [[package]]
@@ -500,7 +541,7 @@ checksum = "de0d48a183585823424a4ce1aa132d174a6a81bd540895822eb4c8373a8e49e8"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.90",
 ]
 
 [[package]]
@@ -728,6 +769,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "lazy_static"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bbd2bcb4c963f2ddae06a2efc7e9f3591312473c50c6685e1f298068316e66fe"
+
+[[package]]
 name = "libc"
 version = "0.2.149"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -857,7 +904,7 @@ dependencies = [
  "phf_shared",
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.90",
  "unicase",
 ]
 
@@ -897,6 +944,16 @@ checksum = "368ba315fb8c5052ab692e68a0eefec6ec57b23a36959c14496f0b0df2c0cecf"
 dependencies = [
  "predicates-core",
  "termtree",
+]
+
+[[package]]
+name = "prettyplease"
+version = "0.2.25"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "64d1ec885c64d0457d564db4ec299b2dae3f9c02808b8ad9c3a089c591b18033"
+dependencies = [
+ "proc-macro2",
+ "syn 2.0.90",
 ]
 
 [[package]]
@@ -1022,6 +1079,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "rustversion"
+version = "1.0.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0e819f2bc632f285be6d7cd36e25940d45b2391dd6d9b939e79de557f7014248"
+
+[[package]]
 name = "ryu"
 version = "1.0.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1034,6 +1097,33 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "93fc1dc3aaa9bfed95e02e6eadabb4baf7e3078b0bd1b4d7b6b0b68378900502"
 dependencies = [
  "winapi-util",
+]
+
+[[package]]
+name = "schemafy_core"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2bec29dddcfe60f92f3c0d422707b8b56473983ef0481df8d5236ed3ab8fdf24"
+dependencies = [
+ "serde",
+ "serde_json",
+]
+
+[[package]]
+name = "schemafy_lib"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "af3d87f1df246a9b7e2bfd1f4ee5f88e48b11ef9cfc62e63f0dead255b1a6f5f"
+dependencies = [
+ "Inflector",
+ "proc-macro2",
+ "quote",
+ "schemafy_core",
+ "serde",
+ "serde_derive",
+ "serde_json",
+ "syn 1.0.109",
+ "uriparse",
 ]
 
 [[package]]
@@ -1052,6 +1142,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "serde-sarif"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a38c5e5bbaa10cc256774ea394ad62968c31c0e3c3265f65221e02c87dd1a914"
+dependencies = [
+ "anyhow",
+ "derive_builder",
+ "prettyplease",
+ "proc-macro2",
+ "quote",
+ "schemafy_lib",
+ "serde",
+ "serde_json",
+ "strum",
+ "strum_macros",
+ "syn 2.0.90",
+ "thiserror",
+]
+
+[[package]]
 name = "serde_derive"
 version = "1.0.215"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1059,7 +1169,7 @@ checksum = "ad1e866f866923f252f05c889987993144fb74e722403468a4ebd70c3cd756c0"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.90",
 ]
 
 [[package]]
@@ -1156,15 +1266,39 @@ checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
 
 [[package]]
 name = "strsim"
-version = "0.10.0"
+version = "0.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "73473c0e59e6d5812c5dfe2a064a6444949f089e20eec9a2e5506596494e4623"
+checksum = "7da8b5736845d9f2fcb837ea5d9e2628564b3b043a70948a3f0b778838c5fb4f"
 
 [[package]]
-name = "strsim"
-version = "0.11.0"
+name = "strum"
+version = "0.26.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5ee073c9e4cd00e28217186dbe12796d692868f432bf2e97ee73bed0c56dfa01"
+checksum = "8fec0f0aef304996cf250b31b5a10dee7980c85da9d759361292b8bca5a18f06"
+
+[[package]]
+name = "strum_macros"
+version = "0.26.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4c6bee85a5a24955dc440386795aa378cd9cf82acd5f764469152d2270e581be"
+dependencies = [
+ "heck",
+ "proc-macro2",
+ "quote",
+ "rustversion",
+ "syn 2.0.90",
+]
+
+[[package]]
+name = "syn"
+version = "1.0.109"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "72b64191b275b66ffe2469e8af2c1cfe3bafa67b529ead792a6d0160888b4237"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "unicode-ident",
+]
 
 [[package]]
 name = "syn"
@@ -1205,6 +1339,26 @@ name = "termtree"
 version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3369f5ac52d5eb6ab48c6b4ffdc8efbcad6b89c765749064ba298f2c68a16a76"
+
+[[package]]
+name = "thiserror"
+version = "1.0.69"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b6aaf5339b578ea85b50e080feb250a3e8ae8cfcdff9a461c9ec2904bc923f52"
+dependencies = [
+ "thiserror-impl",
+]
+
+[[package]]
+name = "thiserror-impl"
+version = "1.0.69"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4fee6c4efc90059e10f81e6d42c60a18f76588c3d74cb83a0b242a2b6c7504c1"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.90",
+]
 
 [[package]]
 name = "thread_local"
@@ -1317,6 +1471,7 @@ dependencies = [
  "proc-exit",
  "regex",
  "serde",
+ "serde-sarif",
  "serde_json",
  "serde_regex",
  "snapbox",
@@ -1432,6 +1587,16 @@ name = "unicode-xid"
 version = "0.2.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ebc1c04c71510c7f702b52b7c350734c9ff1295c464a03335b00bb84fc54f853"
+
+[[package]]
+name = "uriparse"
+version = "0.6.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0200d0fc04d809396c2ad43f3c95da3582a2556eba8d453c1087f4120ee352ff"
+dependencies = [
+ "fnv",
+ "lazy_static",
+]
 
 [[package]]
 name = "utf8parse"
@@ -1702,5 +1867,5 @@ checksum = "9ce1b18ccd8e73a9321186f97e46f9f04b778851177567b1975109d26a08d2a6"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.90",
 ]
--- a/crates/typos-cli/Cargo.toml
+++ b/crates/typos-cli/Cargo.toml
@@ -76,6 +76,7 @@ colorchoice-clap = "1.0.3"
 serde_regex = "1.1.0"
 regex = "1.10.4"
 encoding_rs = "0.8.34"
+serde-sarif = "0.5.0"
 
 [dev-dependencies]
 assert_fs = "1.1"
--- a/crates/typos-cli/src/bin/typos-cli/main.rs
+++ b/crates/typos-cli/src/bin/typos-cli/main.rs
@@ -8,6 +8,8 @@ mod report;
 
 use proc_exit::prelude::*;
 
+use typos_cli::report::Report;
+
 fn main() {
     human_panic::setup_panic!();
     let result = run();
@@ -187,6 +189,13 @@ fn run_checks(args: &args::Args) -> proc_exit::ExitResult {
         None => None,
     };
 
+    // HACK: Diff doesn't handle mixing content
+    let global_reporter = if args.diff {
+        Box::new(report::PrintSilent)
+    } else {
+        args.format.reporter()
+    };
+
     // Note: file_list and args.path are mutually exclusive, enforced by clap
     'path: for path in file_list.as_ref().unwrap_or(&args.path) {
         // Note paths are passed through stdin, `-` is treated like a normal path
@@ -272,14 +281,8 @@ fn run_checks(args: &args::Args) -> proc_exit::ExitResult {
             });
         }
 
-        // HACK: Diff doesn't handle mixing content
-        let output_reporter = if args.diff {
-            Box::new(report::PrintSilent)
-        } else {
-            args.format.reporter()
-        };
-        let status_reporter = report::MessageStatus::new(output_reporter.as_ref());
-        let reporter: &dyn typos_cli::report::Report = &status_reporter;
+        let status_reporter = report::MessageStatus::new(global_reporter.as_ref());
+        let reporter: &dyn Report = &status_reporter;
 
         let selected_checks: &dyn typos_cli::file::FileChecker = if args.files {
             &typos_cli::file::FoundFiles
@@ -331,6 +334,11 @@ fn run_checks(args: &args::Args) -> proc_exit::ExitResult {
         if status_reporter.errors_found() {
             errors_found = true;
         }
+    }
+
+    if let Err(err) = global_reporter.generate_final_result() {
+        errors_found = true;
+        log::error!("could not render end-report: {}", err);
     }
 
     if errors_found {
--- a/crates/typos-cli/src/bin/typos-cli/report.rs
+++ b/crates/typos-cli/src/bin/typos-cli/report.rs
@@ -1,12 +1,15 @@
 #![allow(clippy::needless_update)]
 
 use std::io::Write as _;
-use std::sync::atomic;
+use std::sync::{atomic, Mutex};
 
 use anstream::stdout;
-use unicode_width::UnicodeWidthStr;
-
+use serde_sarif::sarif;
+use serde_sarif::sarif::{
+    ArtifactChangeBuilder, ArtifactContentBuilder, FixBuilder, ReplacementBuilder,
+};
 use typos_cli::report::{Context, Message, Report, Typo};
+use unicode_width::UnicodeWidthStr;
 
 const ERROR: anstyle::Style = anstyle::AnsiColor::BrightRed.on_default();
 const INFO: anstyle::Style = anstyle::AnsiColor::BrightBlue.on_default();
@@ -45,6 +48,10 @@ impl Report for MessageStatus<'_> {
             self.errors_found.store(true, atomic::Ordering::Relaxed);
         }
         self.reporter.report(msg)
+    }
+
+    fn generate_final_result(&self) -> Result<(), std::io::Error> {
+        self.reporter.generate_final_result()
     }
 }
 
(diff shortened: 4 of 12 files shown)


crates/typos-cli/src/bin/typos-cli/report.rs
```diff
@@ -1,12 +1,15 @@
...
-use std::io::Write as _;
-use std::sync::atomic;
+use std::cell::RefCell;
+use std::io::{Error, Write as _};
+use std::sync::{atomic, Mutex};
 
 use anstream::stdout;
+use serde_sarif::sarif;
 use unicode_width::UnicodeWidthStr;
+use serde_sarif::sarif::{MessageBuilder, ResultLevel};
```
epage, line 10: nit: Generally, I prefer to minimize what types get brought in

crates/typos-cli/src/bin/typos-cli/main.rs
```diff
@@ -316,6 +317,8 @@ fn run_checks(args: &args::Args) -> proc_exit::ExitResult {
         if status_reporter.errors_found() {
             errors_found = true;
         }
+        
+        status_reporter.finalize().unwrap();
```
epage, line 321: We need to find something better to do than `unwrap`

crates/typos-cli/src/bin/typos-cli/report.rs
```diff
@@ -280,6 +287,132 @@ impl Report for PrintJson {
...
+    fn report(&self, msg: Message<'_>) -> Result<(), std::io::Error> {
+        match &msg {
+            Message::BinaryFile(_) => {
+                // todo: implement artifacts for sarif
+            }
+            Message::FileType(_) => {
+                // todo: implement artifacts for sarif
+            }
+            Message::File(_) => {
+                // todo: implement artifacts for sarif
```
epage, line 315: todos will need to be resolved in some what before merging

crates/typos-cli/src/bin/typos-cli/report.rs
```diff
@@ -280,6 +287,132 @@ impl Report for PrintJson {
...
+                    typos::Status::Invalid => {
+                        format!("`{}` is disallowed", msg.typo)
+                    }
+                    typos::Status::Corrections(corrections) => {
+                        format!("`{}` should be {}", msg.typo, itertools::join(
+                            corrections.iter().map(|s| format!("`{}`", s)),
+                            ", ",
+                        ))
+                    }
+                };
```
epage, line 337: nit: seems like this should go first to avoid doing stuff in the presence of an early-return

crates/typos-cli/src/bin/typos-cli/report.rs
```diff
@@ -280,6 +287,132 @@ impl Report for PrintJson {
...
+
+                let location = sarif::LocationBuilder::default()
+                    .physical_location(sarif::PhysicalLocationBuilder::default()
+                        .artifact_location(
+                            sarif::ArtifactLocationBuilder::default()
+                                .uri(msg.context.as_ref().map_or("", |context| match context {
+                                    Context::File(context) => context.path.to_str().unwrap_or(""),
+                                    _ => "",
+                                }))
+                                .build().unwrap()
```
epage, line 347: We need to evaluate what should be done with these unwraps

crates/typos-cli/src/bin/typos-cli/report.rs
```diff
@@ -280,6 +287,132 @@ impl Report for PrintJson {
...
+                        )
+                        .region(
+                            sarif::RegionBuilder::default()
+                                .start_line(line_num as i64)
+                                .start_column(column_number as i64)
+                                .end_column((column_number + msg.typo.len()) as i64)
```
epage, line 353: `typo.len()` might not be an appropriate `end_column`.  Since we used `unicode_segmentation::UnicodeSegmentation::graphemes` for start column, we should probably do the same to `msg.typo`.

Granted, all of this depends on what SARIF's definition of a "column" is (bytes, code points, graphemes)

crates/typos-cli/src/bin/typos-cli/report.rs
```diff
@@ -280,6 +287,132 @@ impl Report for PrintJson {
     }
 }
 
+#[derive(Debug)]
+pub(crate) struct PrintSarif {
+    results: Mutex<RefCell<Vec<sarif::Result>>>,
+    error: Mutex<RefCell<Option<String>>>,
```
epage, line 293: Should `error` also be a `Vec`?

crates/typos-cli/src/bin/typos-cli/report.rs
```diff
@@ -280,6 +287,132 @@ impl Report for PrintJson {
...
+            run_builder.invocations(vec![
+                sarif::InvocationBuilder::default()
+                    .process_start_failure_message(error.to_string())
+                    .build().unwrap()
+            ]);
+        }
+
+        let sarif = sarif_builder.build().unwrap();
+
+        write!(stdout().lock(), "{}", serde_json::to_string(&sarif).unwrap())?;
```
epage, line 410: `to_string` or `to_string_pretty`?  Which does sarif expect?

crates/typos-cli/src/bin/typos-cli/report.rs
```diff
@@ -348,7 +481,7 @@ mod tests {
     fn test_calculate_visible_column_width_zwj_sequences() {
         let zwj_sequences = [
             "😵‍💫", /* U+1F635 U+200D U+1F4AB */
-            "👁️‍🗨️",   /* U+1F441 U+FE0F U+200D U+1F5E8 U+FE0F */
+            "👁️‍🗨️", /* U+1F441 U+FE0F U+200D U+1F5E8 U+FE0F */
```
epage, line 484: There are a lot of these small formatting changes.  At least move these to a separate commit

crates/typos-cli/src/report.rs
```diff
@@ -10,11 +10,6 @@ pub trait Report: Send + Sync {
     }
 }
 
-/// A reporter that needs to finalize the report.
```
epage: Please squash into the previous commit

crates/typos-cli/src/bin/typos-cli/report.rs
```diff
@@ -318,10 +318,10 @@ impl Report for PrintSarif {
                 let start = String::from_utf8_lossy(&msg.buffer[0..msg.byte_offset]);
                 let column_number =
                     unicode_segmentation::UnicodeSegmentation::graphemes(start.as_ref(), true).count() + 1;
-                let line_num = msg.context.as_ref().map_or(1, |context| match context {
+                let line_num = msg.context.as_ref().map(|context| match context {
```
epage, line 321: Please squash

crates/typos-cli/Cargo.toml
```diff
@@ -77,6 +77,7 @@ colorchoice-clap = "1.0.3"
 serde_regex = "1.1.0"
 regex = "1.10.4"
 encoding_rs = "0.8.34"
+serde-sarif = "0.4.2"
```
epage, line 80: The docs have this warning:

> [WARNING: VERY UNSTABLE (EARLY IMPLEMENTATION)](https://docs.rs/serde-sarif/latest/serde_sarif/#warning-very-unstable-early-implementation)

crates/typos-cli/src/bin/typos-cli/report.rs
Zxilly: i just have no idea

crates/typos-cli/src/bin/typos-cli/report.rs
Zxilly: the default formmater did this, any rules i can follow？

crates/typos-cli/Cargo.toml
Zxilly: https://github.com/psastras/sarif-rs/commit/9b4adbe37359bbc11a338e4e1d789d81f26ce535

This was added three years ago. As I only used the generated code from a json schema, I think this won't be a problem.

crates/typos-cli/src/bin/typos-cli/report.rs
Zxilly: I don't know how to do this more `rust`. even I check it earlier, the match statement still needs a branch for that. use unreachable!？

crates/typos-cli/src/bin/typos-cli/report.rs
Zxilly: sarif basically a json file. I think both is ok as it only requires machine-readable

crates/typos-cli/src/bin/typos-cli/main.rs
Zxilly: I just have no idea, ? can not cast error automatic, should i attach a map_err for everyone?

Zxilly:
@epage Could you please approve the workflow run to see if anything broken by this?

crates/typos-cli/src/bin/typos-cli/main.rs
epage: Why is this marked as resolved when nothing has changed?

crates/typos-cli/src/bin/typos-cli/report.rs
```diff
@@ -280,6 +286,148 @@ impl Report for PrintJson {
...
+
+                self.results.lock().unwrap().borrow_mut().push(result);
+            }
+            Message::Parse(msg) => {
+                self.error.lock().unwrap().borrow_mut().push(msg.data.to_owned());
+            }
+            Message::Error(msg) => {
+                self.error.lock().unwrap().borrow_mut().push(msg.msg.clone());
+            }
+            _ => {}
```
epage, line 376: Please enumerate all known values so it is clear what is happening

crates/typos-cli/src/bin/typos-cli/report.rs
epage: If its truly unreachable, sure, but that should be verified and likely should have a message with it to say why you know its unreachable.

If it isn't truely unreachable, then please move this block before `let start`.

crates/typos-cli/src/bin/typos-cli/report.rs
```diff
@@ -280,6 +286,148 @@ impl Report for PrintJson {
...
+impl Default for PrintSarif {
+    fn default() -> Self {
+        Self {
+            results: Mutex::new(RefCell::new(Vec::new())),
+            error: Mutex::new(RefCell::new(Vec::new())),
+        }
+    }
+}
+
+fn error_to_io_error(error: Box<dyn std::error::Error>) -> std::io::Error {
```
epage, line 304: Please move a helper like this to down below the `impl`.  At least for me, `impl`s should come right after their type

epage, line 304: You can skip the `e.into()`s by making this
```rust
fn error_to_io_error(error: impl Into<Box<dyn std::error::Error>>) -> std::io::Error {
    let error = error.into();
    ...
}
```
Though if all you are doing is `to_string`, you can instead do
```rust
fn error_to_io_error(error: impl std::fmt::Display) -> std::io::Error {
    ...
}
```

crates/typos-cli/src/bin/typos-cli/report.rs
```diff
@@ -280,6 +286,148 @@ impl Report for PrintJson {
...
+
+
+                let result = sarif::ResultBuilder::default()
+                    .level(sarif::ResultLevel::Error.to_string())
+                    .message(sarif::MessageBuilder::default()
+                        .markdown(message)
+                        .build().map_err(|e| { error_to_io_error(e.into()) })?)
+                    .locations(vec![location])
+                    .build().map_err(|e| { error_to_io_error(e.into()) })?;
+
```
epage, line 367: One way to simplify these `map_err`s is to pull the logic out into a helper that returns the error from sarif.

crates/typos-cli/src/bin/typos-cli/report.rs
```diff
@@ -280,6 +286,148 @@ impl Report for PrintJson {
...
+                                    _ => "",
+                                }).unwrap_or(""))
+                                .build().map_err(|e| { error_to_io_error(e.into()) })?
+                        )
+                        .region(
+                            sarif::RegionBuilder::default()
+                                .start_line(line_num as i64)
+                                .start_column(column_start as i64)
+                                .end_column(column_end as i64)
+                                .build().map_err(|e| { error_to_io_error(e.into()) })?
```
epage, line 354: In 3.30.2 Text regions

> A text region MAY be specified in two ways:
> * By means of the “line/column” properties startLine (§3.30.5), startColumn (§3.30.6), endLine (§3.30.7), and endColumn (§3.30.8).

Its a **MAY** but let's specify `endLine` as well

epage: though looks like the default is `startLine`

>  "endLine" is absent, and so defaults to "startLine", which in this example is 1 (see §3.30.7).

crates/typos-cli/src/bin/typos-cli/report.rs
```diff
@@ -280,6 +286,148 @@ impl Report for PrintJson {
...
+
+impl Report for PrintSarif {
+    fn report(&self, msg: Message<'_>) -> Result<(), std::io::Error> {
+        match &msg {
+            Message::Typo(msg) => {
+                let start = String::from_utf8_lossy(&msg.buffer[0..msg.byte_offset]);
+                let column_start =
+                    unicode_segmentation::UnicodeSegmentation::graphemes(start.as_ref(), true).count() + 1;
+                let column_end =
+                    unicode_segmentation::UnicodeSegmentation::graphemes(msg.typo, true).count() + column_start;
```
epage, line 316: 1. Looking at the spec, we should be reporting code points, not graphemes
2. We need to specify the column unit

> If a SARIF producer processes text artifacts and theRun.results (§3.14.23) is non-empty, the run object SHALL contain a property named columnKind whose value is a string that specifies the unit in which the analysis tool measures columns. If a SARIF producer processes text artifacts and theRun.results is empty, columnKind MAY be present.
> 
> ...
>
> * "unicodeCodePoints": Each Unicode code point (abstract character) is considered to occupy one column. This means that even a character that is represented in UTF-16 by a surrogate pair is considered to occupy one column.

crates/typos-cli/src/bin/typos-cli/report.rs
epage: Started a separate comment on the units of `column`

crates/typos-cli/src/bin/typos-cli/report.rs
epage: If SARIF uses json and not jsonlines, let's go with `to_string_pretty` to make this easier for humans

crates/typos-cli/src/bin/typos-cli/report.rs
epage: Why is this marked as resolved when it isn't resolved?

crates/typos-cli/Cargo.toml
epage: The warning is still there and I would like more concrete detail (and not supposition) on why its there.

Zxilly: Then maybe an issue confirmation should be opened in the upstream repository?

crates/typos-cli/src/bin/typos-cli/report.rs
Zxilly: I just restored the original format. But apparently this causes the format check to fail in CI

crates/typos-cli/src/bin/typos-cli/main.rs
Zxilly: I forgeted to update this one, will fixed in the new patch.

crates/typos-cli/src/bin/typos-cli/report.rs
Zxilly: https://github.com/crate-ci/typos/pull/1048

crates/typos-cli/src/bin/typos-cli/report.rs
Zxilly: updated

crates/typos-cli/src/bin/typos-cli/report.rs
Zxilly: I set the columnKind as `unicodeCodePoints`. So the `graphemes().count()` value should represents the code point correctly.

crates/typos-cli/src/bin/typos-cli/report.rs
Zxilly: So let's leave it as it is.

crates/typos-cli/Cargo.toml
Zxilly: https://github.com/psastras/sarif-rs/issues/646

But the author doesn't seem to be active and also takes a long time to respond.

Zxilly: The author set it to stable now.

Zxilly:
@epage I've resolved the conflict, can you take a look at this?

Zxilly:
@epage hello, any updates?

Cargo.lock
```diff
@@ -14,9 +14,9 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.21.0"
+version = "0.22.0"
```
epage, line 17: Why was this lockfile update done?

crates/typos-cli/Cargo.toml
```diff
@@ -77,7 +77,7 @@ colorchoice-clap = "1.0.3"
 serde_regex = "1.1.0"
 regex = "1.10.4"
 encoding_rs = "0.8.34"
-serde-sarif = "0.4.2"
+serde-sarif = "0.5.0"
```
epage, line 80: Why was this upgrade done?  Why wasn't it in the prior commit?

crates/typos-cli/src/bin/typos-cli/report.rs
epage: Ok

crates/typos-cli/src/bin/typos-cli/main.rs
epage: > log::error!("Error finalizing: {}", err);

Is not an appropriate way of handling this

crates/typos-cli/src/bin/typos-cli/report.rs
epage: Ok

crates/typos-cli/src/bin/typos-cli/report.rs
```diff
@@ -280,6 +286,155 @@ impl Report for PrintJson {
...
+                        format!("`{}` is disallowed", msg.typo)
+                    }
+                    typos::Status::Corrections(corrections) => {
+                        format!("`{}` should be {}", msg.typo, itertools::join(
+                            corrections.iter().map(|s| format!("`{}`", s)),
+                            ", ",
+                        ))
+                    }
+                    typos::Status::Valid => { unreachable!("has been checked earlier") }
+                };
```
epage, line 334: Why are we doing two checks on `msg.corrections`?  These should be merged

crates/typos-cli/src/bin/typos-cli/report.rs
epage: I don't think this is resolved.

crates/typos-cli/src/bin/typos-cli/report.rs
```diff
@@ -280,6 +286,155 @@ impl Report for PrintJson {
...
+                let error = error.unwrap().as_ref().unwrap_err();
+                return Err(std::io::Error::new(error.kind(), error.to_string()));
+            }
+            
+            let invocations: Vec<Invocation> = invocations.into_iter().map(|x| x.unwrap()).collect();
+
+            run_builder.invocations(
+                invocations
+            );
+        }
```
epage, line 424: It looks like this 
1. Converts the error
2. Checks for the error
3. Unwraps because the error is handled

Generally avoid checking for an error and then unwrapping.  Most likely, the culprit for this is over-reliance on combinators.

crates/typos-cli/src/bin/typos-cli/report.rs
```diff
@@ -280,6 +286,155 @@ impl Report for PrintJson {
...
+
+        let sarif = sarif_builder.build().map_err(error_to_io_error)?;
+
+        write!(stdout().lock(), "{}", serde_json::to_string_pretty(&sarif)?)?;
```
epage, line 428: Shouldn't this be [`to_writer_pretty`](https://docs.rs/serde_json/latest/serde_json/fn.to_writer_pretty.html)?

crates/typos-cli/src/bin/typos-cli/report.rs
epage: Ok

crates/typos-cli/src/bin/typos-cli/report.rs
epage: Ok

crates/typos-cli/src/report.rs
epage: Ok

crates/typos-cli/Cargo.toml
epage: Ok

crates/typos-cli/src/bin/typos-cli/report.rs
epage: Ok

crates/typos-cli/src/bin/typos-cli/report.rs
epage: This isn't done.  In case there was confusion, this was about splitting it out like
```rust


impl Report for PrintSarif {
    fn report(&self, msg: Message<'_>) -> Result<(), std::io::Error> {
        report_sarif(msg).map_err(error_to_io_error)
    }
}

fn report_sarif(&self, msg: Message<'_>) -> Result<(), _> {
    // ...
}
```

crates/typos-cli/src/bin/typos-cli/report.rs
epage: Ok

crates/typos-cli/src/bin/typos-cli/report.rs
epage: Isn't a `char` a code point while a grapheme can be made of multiple multiple `char`s?

crates/typos-cli/Cargo.toml
Zxilly: Please take a loot at https://github.com/psastras/sarif-rs/issues/646.

epage: That doesn't answer either question on its own.

crates/typos-cli/src/bin/typos-cli/report.rs
```diff
@@ -280,6 +286,155 @@ impl Report for PrintJson {
     }
 }
 
+#[derive(Debug)]
+pub(crate) struct PrintSarif {
+    results: Mutex<RefCell<Vec<sarif::Result>>>,
+    error: Mutex<RefCell<Vec<String>>>,
```
epage, line 292: Why do we need `RefCell` when we have a `Mutex`?

crates/typos-cli/src/bin/typos-cli/main.rs
```diff
@@ -316,6 +317,13 @@ fn run_checks(args: &args::Args) -> proc_exit::ExitResult {
         if status_reporter.errors_found() {
             errors_found = true;
         }
+        
+        match status_reporter.finalize() {
+            Ok(_) => {}
+            Err(err) => {
+                log::error!("Error finalizing: {}", err);
```
epage, line 324: `Error` would be redundant with log showing `ERROR`.  `finalizing` has no user-facing meaning in this context

crates/typos-cli/src/bin/typos-cli/main.rs
```diff
@@ -316,6 +317,13 @@ fn run_checks(args: &args::Args) -> proc_exit::ExitResult {
         if status_reporter.errors_found() {
             errors_found = true;
         }
+        
+        match status_reporter.finalize() {
```
epage, line 321: still not thrilled with `finalize`, both the role and the name.  That is very much a C#-ism.

crates/typos-cli/Cargo.toml
Zxilly: This version makes no code changes, bumps the version number, updates some dependencies, and then removes the experimental warning.
The bumping of the version was done to address the issue of https://github.com/crate-ci/typos/pull/1047#discussion_r1659499185

Cargo.lock
Zxilly: `cargo generate-lockfile` did this automatically.

crates/typos-cli/Cargo.toml
epage: Or in other words...

> Why was this upgrade done? 

To pull in a version that was listed as unstable. This had no other affect as any breaking changes in that release do not affect us.

>  Why wasn't it in the prior commit?

*This part is still left unanswered*

Cargo.lock
epage: Why was that run?

crates/typos-cli/src/bin/typos-cli/report.rs
Zxilly: to resolve https://github.com/crate-ci/typos/pull/1047#discussion_r1659498218

crates/typos-cli/src/bin/typos-cli/report.rs
Zxilly: Since `report` in `Report` has a immutable `self`, without `RefCell` I can write to `results` and `error` in the `report`.

crates/typos-cli/src/bin/typos-cli/report.rs
Zxilly: How about https://github.com/crate-ci/typos/pull/1047/commits/5ad1fb9e130db2d5ebc9c0bc5699413bbf45ea7b ?

crates/typos-cli/src/bin/typos-cli/main.rs
Zxilly: updated

crates/typos-cli/Cargo.toml
Zxilly: At the time I wrote the initial code for this PR, that version had not yet been released. Since you mentioned that the dependency was unstable, I went and asked the author, who bumped up the version number, and I followed the update as a result.

Cargo.lock
Zxilly: An update on the master branch caused a CONFLICT in the lockfile in this PR, so I recreated it.

crates/typos-cli/src/bin/typos-cli/report.rs
Zxilly: done

crates/typos-cli/src/bin/typos-cli/report.rs
Zxilly: Meaning implement this only for sarif, not extract a trait that requires context?

crates/typos-cli/src/bin/typos-cli/report.rs
epage: I recreated this in the playground:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4f989848d30f923f7c3ccc3b5d3c6ac6

I then removed `RefCell` and it still works:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fad6216927b55e6b98435cba8209abb9

`Mutex` fills a similar role to `RefCell` except its thread-safe.

crates/typos-cli/src/bin/typos-cli/report.rs
epage: Unsure what is meant by either option of the question.

I'm just referring to extracting the function body that has a sarif-focued error type so you can use `?` without having to `map_err` every time, only doing it in the wraper call that handles the trait impl.

Cargo.lock
epage: Please don't throwaway existing lockfiles.  If nothing else, this cases a lot of unnecessary churn to a PR.  A way to resolve conflicts like this is `cargo checkout master -- Cargo.lock && cargo check`.

crates/typos-cli/Cargo.toml
epage: So this is an update for the reviewer and has no distinct meaning as a separate commit?  Then that should be squashed like the other cases that have been squashed before.

crates/typos-cli/src/bin/typos-cli/report.rs
epage: Thats not what was being asked for there.  That was asking to move the entire `match` up earlier in the function so we would return early on `Valid`, before doing some of the processing that is currently done before `match`.

crates/typos-cli/src/bin/typos-cli/report.rs
epage: Thanks, that cleans it up

crates/typos-cli/src/bin/typos-cli/main.rs
epage: `generate final result` still doesn't have much meaning to the user and reads oddly within context of an error.

Likely we should directly print the error and have `error_to_io_error` change the error message to `format!("failed to generate SARIF output: {err}")`

crates/typos-cli/src/bin/typos-cli/report.rs
Zxilly: As https://www.unicode.org/reports/tr29/#Grapheme_Cluster_Boundaries described, `UnicodeSegmentation::graphemes` should mean to be `unicodeCodePoints` in `SARIF`, because they all represent a user-readable character, even though the underlying layer consists of multiple code points.

> SARIF: This means that even a character that is represented in UTF-16 by a surrogate pair is considered to occupy one column.

> Unicode: For example, “G” + grave-accent is a user-perceived character: users think of it as a single character, yet is actually represented by two Unicode code points. These user-perceived characters are approximated by what is called a grapheme cluster, which can be determined programmatically.

crates/typos-cli/src/bin/typos-cli/report.rs
Zxilly: updated

crates/typos-cli/src/bin/typos-cli/report.rs
Zxilly: Well it did work, I know so little about rust

crates/typos-cli/src/bin/typos-cli/main.rs
Zxilly: updated

crates/typos-cli/src/bin/typos-cli/report.rs
Zxilly: I see what you mean. I'll update the code now.

crates/typos-cli/src/bin/typos-cli/report.rs
epage: > SARIF: This means that even a character that is represented in UTF-16 by a surrogate pair is considered to occupy one column.

From my understanding, UTF-16 surrogate pairs are a way to represent code points that can't be represented otherwise.  Grapheme's are more than just surrogate pairs.

crates/typos-cli/src/bin/typos-cli/report.rs
```diff
@@ -154,7 +158,7 @@ fn print_brief_correction(msg: &Typo<'_>) -> Result<(), std::io::Error> {
                 msg.typo,
                 itertools::join(
                     corrections.iter().map(|s| format!("`{good}{}{reset}`", s)),
-                    ", "
+                    ", ",
```
epage, line 161: FYI unrelated formatting got inserted

crates/typos-cli/src/bin/typos-cli/main.rs
```diff
@@ -272,7 +272,7 @@ fn run_checks(args: &args::Args) -> proc_exit::ExitResult {
             args.format.reporter()
         };
         let status_reporter = report::MessageStatus::new(output_reporter.as_ref());
-        let reporter: &dyn typos_cli::report::Report = &status_reporter;
+        let reporter: &dyn Report = &status_reporter;
```
epage, line 275: Unrelated change?

Cargo.lock
Zxilly: done

crates/typos-cli/Cargo.toml
Zxilly: squashed

crates/typos-cli/src/bin/typos-cli/main.rs
```diff
@@ -316,6 +317,11 @@ fn run_checks(args: &args::Args) -> proc_exit::ExitResult {
         if status_reporter.errors_found() {
             errors_found = true;
         }
+
+        if let Err(err) = status_reporter.generate_final_result() {
+            log::error!("generate final result: {}", err);
+            return proc_exit::Code::FAILURE.ok();
+        }
```
epage, line 324: This has us printing one SARIF json object per argument.

This should either be moved out of the loop or this is jsonlines and we shouldn't use `pretty`

crates/typos-cli/src/bin/typos-cli/main.rs
```diff
@@ -316,6 +317,11 @@ fn run_checks(args: &args::Args) -> proc_exit::ExitResult {
         if status_reporter.errors_found() {
             errors_found = true;
         }
+
+        if let Err(err) = status_reporter.generate_final_result() {
+            log::error!("generate final result: {}", err);
+            return proc_exit::Code::FAILURE.ok();
```
epage, line 323: Why was the early return added?  Should we instead set `errors_found`?

crates/typos-cli/src/bin/typos-cli/report.rs
epage: ok

crates/typos-cli/src/bin/typos-cli/report.rs
epage: ok

crates/typos-cli/Cargo.toml
epage: ok

Cargo.lock
epage: ok

crates/typos-cli/src/bin/typos-cli/report.rs
epage: ok

crates/typos-cli/src/bin/typos-cli/report.rs
epage: ok

crates/typos-cli/src/bin/typos-cli/main.rs
epage: Not seeing this updated

crates/typos-cli/src/bin/typos-cli/report.rs
Zxilly: updated

epage: ok

Zxilly:
I squashed the commits again to clean up changes not related to this PR.

crates/typos-cli/src/bin/typos-cli/main.rs
Zxilly: updated

crates/typos-cli/src/bin/typos-cli/main.rs
Zxilly: That's too bad I didn't notice that typos created a reporter for each file. i need to keep elevating the status of the reporter to make sure that the sarif file contains information from all the files.

crates/typos-cli/src/bin/typos-cli/report.rs
Zxilly: removed

crates/typos-cli/src/bin/typos-cli/main.rs
Zxilly: removed

crates/typos-cli/src/bin/typos-cli/main.rs
Zxilly: please take a look at https://github.com/crate-ci/typos/pull/1047/commits/7a834d04ec551250395aa1e72d8c59314e3f322c

crates/typos-cli/src/bin/typos-cli/report.rs
Zxilly: I switch to `utf16CodeUnits`

Zxilly:
The conflict has been resolved.
Note that the winnow upgrade deprecated a certain API that causes linter to display a warning now, which is unrelated to the changes made by this PR.

Zxilly:
@epage Can we move forward with this patch?

Zxilly:
@epage I have resolved the conflict.

epage:
Sorry for the delays as I am slowly working through [my backlog](https://hachyderm.io/@epage).

It would help if the commits were cleaned up to present this how you want it reviewed and merged.

It would also be good for us to have a test for this, likely among our end-to-end tests.

Zxilly:
I suppose I could write some integration tests, but I'm not quite sure where the old tests are located, is there any architecture I could reuse?

Zxilly:
I have added a test

epage:
Sorry that I've fallen behind and not been able to get back to this yet.

I did want to quickly note that I just noticed the `fix` element
https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141791131

I wonder if github has integration for that so people get diffs in the post with the possibility of making a commit right from the review.  

Note: I would not block merging this PR on getting that to work

crates/typos-cli/src/bin/typos-cli/report.rs
```diff
@@ -435,6 +435,9 @@ impl PrintSarif {
 
             run_builder.invocations(invocations.unwrap());
         }
+        
+        let run = run_builder.build()?;
+        sarif_builder.runs(vec![run]);
```
epage, line 440: this seems to be in the wrong commit

crates/typos-cli/src/bin/typos-cli/main.rs
```diff
@@ -187,6 +189,11 @@ fn run_checks(args: &args::Args) -> proc_exit::ExitResult {
         None => None,
     };
 
+    let global_reporter = match args.format {
```
epage, line 192: Why was `global_reporter` added?  This requires knowledge of the reporter's implementation to know whether it should be global or not which makes this more brittle

crates/typos-cli/tests/cmd/stdin-failure-sarif.toml
```diff
@@ -0,0 +1,46 @@
+bin.name = "typos"
+args = "--format sarif -"
+status.code = 2
+stdin = '''
+Apropriate world
+'''
```
epage, line 6: Can you update the test so the invalid word is in the middle of content?

crates/typos-cli/src/bin/typos-cli/report.rs
```diff
@@ -280,6 +285,172 @@ impl Report for PrintJson {
...
+                let column_start = start.encode_utf16().count() + 1;
+                let column_end = msg.typo.encode_utf16().count() + column_start;
+                let line_num = msg
+                    .context
+                    .as_ref()
+                    .map(|context| match context {
+                        Context::File(context) => context.line_num,
+                        _ => 1,
+                    })
+                    .unwrap_or(1);
```
epage, line 343: I'm a little concerned about how we create locations.

Could we mirror what `clippy-sarif` does?
See https://github.com/psastras/sarif-rs/blob/main/serde-sarif/src/converters/clippy.rs

(note that also has suggestion support)

crates/typos-cli/src/bin/typos-cli/main.rs
Zxilly: Because all previous reporter require only partial state, the output can be written as soon as any event arrives. But sarif must wait until all processing is complete, then collect the output and aggregate it into a single file.

epage: Maybe put another way, why can't they all be global?

Zxilly: I think it's possible, but it would require staging all events while processing, adding an unnecessary memory footprint. I can refactor the old reporter if you think simplicity in code is more important.

crates/typos-cli/src/bin/typos-cli/report.rs
Zxilly: suggestion support has been implement

crates/typos-cli/src/bin/typos-cli/main.rs
epage: Not seeing why we'd need to do that.  I just pushed my suggested change and it seems to work.

epage:
- Switched to code points
- Added more tests
- Improve handling of file name typos

epage:
@Zxilly mind doing a pass over of my changes?

Zxilly:
Look good to me

PR #1048, opened 2024-06-29
style: format with `cargo fmt --all`

Zxilly:
https://github.com/crate-ci/typos/pull/1047#discussion_r1659498970_


epage:
What version of `cargo fmt` are you using?  I can't reproduce this.

Zxilly:
I use
```
cargo version
cargo 1.81.0-nightly (4ed7bee47 2024-06-25)
```

Zxilly:
Considering the problem is next to unicode characters, maybe it's the OS? I'm using Windows 11

epage:
I just checked and it is stable vs nightly.  As we enforce stable, we're going to stick with that

Zxilly:
I've also successfully reproduced it on stable.

```
rustup default
stable-x86_64-pc-windows-msvc (default)
```

Zxilly:
Also reproduced on wsl
image

PR #1059, opened 2024-07-20
fix the crate install path in Dockerfile

huuff:
The Dockerfile might have been old (from before the project was a workspace) and the path to the installable crate was wrong. 

This also broke the pre-commit docker hook, which was important for me since I couldn't get the default one to work on NixOS.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ ARG DEBIAN_DIST=bullseye
 FROM rust:${DEBIAN_DIST} as builder
 WORKDIR /usr/src/typos
 COPY . .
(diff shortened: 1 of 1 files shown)


huuff:
As a side note, I just realized that maybe the main hook doesn't work because it's configured as python instead of rust. I'd be happy to submit a PR for that one, too.

epage:
> main hook doesn't work because it's configured as python instead of rust. I'd be happy to submit a PR for that one, too.

The main pre-commit hook is correct as it installs via https://pypi.org/project/typos/

PR #1096, opened 2024-09-10
feat(action): Add support for force-exclude option

manic:
When using typos with changed-files, the `extend-exclude` setting in the configuration file is ignored, leading to checks on unwanted directories.

This update ensures target directories are consistently ignored and avoids unnecessary validations.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -25,6 +25,11 @@ inputs:
     required: false
     default: false
 
+  force_exclude:
+    description: "Force exclude of files or patterns"
+    required: false
+    default: false
+
   config:
     description: "Use a custom config file."
     required: false
@@ -42,6 +47,7 @@ runs:
         INPUT_EXTEND_WORDS: ${{ inputs.extend_words }}
         INPUT_ISOLATED: ${{ inputs.isolated }}
         INPUT_WRITE_CHANGES: ${{ inputs.write_changes }}
+        INPUT_FORCE_EXCLUDE: ${{ inputs.force_exclude }}
         INPUT_CONFIG: ${{ inputs.config }}
 
 branding:
--- a/action/entrypoint.sh
+++ b/action/entrypoint.sh
@@ -63,6 +63,11 @@ if [ "${INPUT_WRITE_CHANGES:-false}" == "true" ]; then
     ARGS+=" --write-changes"
 fi
 
+# Force exclude of files or patterns
+if [ "${INPUT_FORCE_EXCLUDE:-false}" == "true" ]; then
+    ARGS+=" --force-exclude"
+fi
+
 # Use a custom configuration file
 if [[ -n "${INPUT_CONFIG:-}" ]]; then
     ARGS+=" --config ${INPUT_CONFIG}"


epage:
As for this change, I wonder if we should unconditionally set it like we do with pre-commit.

> When using typos with changed-files

I'd be curious to see your setup for this

manic:
This is our current `.github/workflows/typos.yml` file with my patch:

```
name: Check typos
on:
  pull_request:
jobs:
  run:
    name: Spell Check with Typos
    runs-on: ubuntu-latest
    steps:
    - name: Checkout Actions Repository
      uses: actions/checkout@v4
    - name: Get changed files
      id: changed-files
      uses: tj-actions/changed-files@v14.4
    - name: Check spelling of file.txt
      if: ${{ steps.changed-files.outputs.all_changed_files != '' }}
      uses: manic/typos@master
      with:
        files: ${{ steps.changed-files.outputs.all_changed_files }}
        config: ./.typos.toml
        force_exclude: true
```

The goal is to only check the spelling for files that were changed in the pull request. This reduces the number of files to process, allowing for quicker feedback.

epage:
>  allowing for quicker feedback.

How much of a time difference does that make?

manic:
@epage Sorry, after double-checking, there’s no noticeable time difference between running typos with a diff versus running it on all files.

However, our repositories include YAML files and settings that contain other languages and base64 strings, which are incorrectly flagged as typos. We need to ignore these to prevent the check from failing every time.

Here are some examples:
```
error: `Iy` should be `It`
  --> ./fixtures/vcr_cassettes/..../update_profile_business_token_invalid_token.yml:155:60
    |
155 |         e_proxy="AcLD8bfE3DYo_IwZea6jVgUFxE1h8SfVCg2Dspz0cDIy2xXX2DsBUa2JUEXRzlF9fO-wtLPs3tjpB_vraUVp",

error: `accout` should be `account`
  --> ./fixtures/vcr_cassettes/..../make_reservation_fail_due_to_more_than_allowed_quota.yml:1171:16
     |
1171 |         name=\"accoutID__c\" nillable=\"true\" minOccurs=\"0\" type=\"xsd:string\"/>\n
     |                ^^^^^^
     |
```

epage:
Sounds like filtering what files are checked masks problems
- This will show up if those files are edited or new files created
- If you update `typos` (or have it auto-update), you won't find any new typos introduced until those files are touched
- Running `typos` locally will fail

Seems like that should be fixed more generally, see https://github.com/crate-ci/typos?tab=readme-ov-file#false-positives.  In particular, our [refernece](https://github.com/crate-ci/typos/blob/master/docs/reference.md) has examples of regexes to create directives for disabling typos within sections of files.

manic:
@epage, I think there might be a slight misunderstanding regarding my patch.

In our GitHub workflow, we are using typos in combination with changed-files to check only the files modified in a PR.
We have also defined an extend-exclude list to ignore certain files or directories, such as the fixtures directory.
However, when a PR contains changes that include files from the fixtures directory, the typos check is still running on those files, despite being on the exclusion list.
After some investigation, we discovered that adding force-exclude is necessary to ensure that the extend-exclude settings work as expected and properly skip those directories.
In short, force-exclude is required to make sure our defined exclusions (like the fixtures directory) are respected during the typo checks.

epage:
If the performance difference is negligable, why then care?   I had thought the statement about YAML files was about why you cared but that appears to not be the case.

PR #1114, opened 2024-10-06
Updates stage names in `.pre-commit-hooks.yaml`.

tobiasraabe:
The stage names were updated in pre-commit v4 and the old names are deprecated.

- https://github.com/pre-commit/pre-commit/releases/tag/v4.0.0
- https://pre-commit.com/#confining-hooks-to-run-at-certain-stages

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -5,7 +5,7 @@
   entry: typos
   args: [--write-changes, --force-exclude]
   types: [text]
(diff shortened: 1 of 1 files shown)


epage:
The deprecation was added 2 days ago.  Whats more important is when the new stage names were added which turns out to be  3.2, see https://github.com/pre-commit/pre-commit/blob/main/CHANGELOG.md#320---2023-03-17

So they have been around for over a year and a half.  Seems reasonable enough of lead time to make this change.

Issue #1106, opened 2024-10-01
Non-critical corrections - October 2024 Edition


epage:
Please list any corrections you noticed `typos` doesn't fix and we'll roll these up into a single PR, making it lighter weight for adding new corrections.

Preferred format (for easy copy/pasting):
````
```
bad1 -> good1
bad2 -> good2
```
````
*(using a code block allows me one-click copy/paste)*

Feel free to update your comment as you find more suggestions for this month to reduce noise and making updating the dictionary easier

Please use dedicated issues for false positives as they likely need more discussion than is intended for this thread.

Related open issues
- #466
- #515
- #592
- #743
- #745
- #795

**Comment by NathanBaulch** [comment]
```
alarams -> alarms
budegets -> budgets
detalis -> details
echosystem -> ecosystem
empyty -> empty
exansion -> expansion
existiong -> existing
fascilitating -> facilitating
idenfier -> identifier
idenitier -> identifier
incuring -> incurring
intergations -> integrations
privisoned -> provisioned
provisoner -> provisioner
reintialize -> reinitialize
reploying -> redeploying
subscibres -> subscribers
subscibresendpoings -> endpoints
sythetics -> synthetics
wedeged -> wedged
```

**Comment by jrosdahl** [comment]
```
balue -> blue, value # not just blue
```

**Comment by kachick** [comment]
```
didin't -> didn't
```

**Comment by joshpetit** [comment]
Is there a thread/format for things that should not be corrected?

For instance, "Seldomly" is marked as an incorrect word

**Comment by ember91** [comment]
whitin -> within

**Comment by akx** [comment]
```
exepect -> expect
assined -> assigned
```

**Comment by epage** [comment]
@joshpetit no, those should be handled in individual issues. Please open one.

**Comment by Veetaha** [comment]
```
meeans -> means
```

**Comment by waywardmonkeys** [comment]
```
metamophosis -> metamorphosis
```

**Comment by overlookmotel** [comment]
```
proeperty -> property
```

**Comment by overlookmotel** [comment]
> `subscibres -> subscribers`

I suggest instead:

```
subscibres -> subscribers, subscribes
subscribres -> subscribers, subscribes
```

**Comment by LeoDog896** [comment]
```
substitte -> substitute
```

**Comment by szepeviktor** [comment]
```
degister -> deregister
```

**Comment by justanotheranonymoususer** [comment]
```
notorization -> notarization
```

szepeviktor:
[...]

PR #1115, opened 2024-10-08
fix(dict): additional fix for password

nicholaskuechler:
Adds correction of `pasasword` to `password`


epage:
Wanted to check if you were aware of #1106 for these being processed in bulk once a month and if this could wait for that.

nicholaskuechler:
Left a comment on the other issue. Thanks.

PR #1119, opened 2024-10-10
feat: add --interactive option to prompt for each change

### LINKED ISSUE #397: Allow interactive usage with prompt to confirm that changes are written
When spellchecking a lager project with lots of history and contributions from all over the world for the first time, I would like to confirm each change before it is written and eventually collect cases to be added to exceptions from automatic spell correction.

Therefore, an interactive flag, that adds a prompt before changes are written would be very useful for initial spell checking.

Bnyro:
The code is probably far from perfect, I'm hoping for some help at improving the code quality in the review :)

closes #397

### DIFF
--- a/crates/typos-cli/src/bin/typos-cli/args.rs
+++ b/crates/typos-cli/src/bin/typos-cli/args.rs
@@ -67,6 +67,10 @@ pub(crate) struct Args {
     #[arg(long, short = 'w', group = "mode", help_heading = "Mode")]
     pub(crate) write_changes: bool,
 
+    /// Prompt for each suggested correction whether to write the fix
+    #[arg(long, short = 'i', group = "mode", help_heading = "Mode")]
+    pub(crate) interactive: bool,
+
     /// Debug: Print each file that would be spellchecked.
     #[arg(long, group = "mode", help_heading = "Mode")]
     pub(crate) files: bool,
--- a/crates/typos-cli/src/bin/typos-cli/main.rs
+++ b/crates/typos-cli/src/bin/typos-cli/main.rs
@@ -32,6 +32,14 @@ fn run() -> proc_exit::ExitResult {
 
     init_logging(args.verbose.log_level());
 
+    // HACK: Ensure the terminal gets reset to a good state if the user hits ctrl-c during a prompt
+    // https://github.com/console-rs/dialoguer/issues/294
+    ctrlc::set_handler(move || {
+        let _ = console::Term::stdout().show_cursor();
+        std::process::exit(130);
+    })
+    .expect("Failed to set handler for Ctrl+C needed to restore terminal defaults after killing the process");
+
     if let Some(output_path) = args.dump_config.as_ref() {
         run_dump_config(&args, output_path)
     } else if args.type_list {
@@ -289,6 +297,8 @@ fn run_checks(args: &args::Args) -> proc_exit::ExitResult {
             &typos_cli::file::Identifiers
         } else if args.words {
             &typos_cli::file::Words
+        } else if args.interactive {
+            &typos_cli::file::Interactive
         } else if args.write_changes {
             &typos_cli::file::FixTypos
         } else if args.diff {
--- a/crates/typos-cli/src/file.rs
+++ b/crates/typos-cli/src/file.rs
@@ -1,4 +1,5 @@
 use bstr::ByteSlice;
+use dialoguer::{Confirm, Select};
 use std::io::Read;
 use std::io::Write;
 
@@ -126,6 +127,87 @@ impl FileChecker for FixTypos {
                         reporter.report(msg.into())?;
                     }
                 }
+                if !fixes.is_empty() {
+                    let new_path = fix_file_name(path, file_name, fixes.into_iter())?;
+                    std::fs::rename(path, new_path)?;
+                }
+            }
+        }
+
+        Ok(())
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct Interactive;
+
+impl FileChecker for Interactive {
+    fn check_file(
+        &self,
+        path: &std::path::Path,
+        explicit: bool,
+        policy: &crate::policy::Policy<'_, '_, '_>,
+        reporter: &dyn report::Report,
+    ) -> Result<(), std::io::Error> {
+        if policy.check_files {
+            let (buffer, content_type) = read_file(path, reporter)?;
+            let bc = buffer.clone();
+            if !explicit && !policy.binary && content_type.is_binary() {
+                let msg = report::BinaryFile { path };
+                reporter.report(msg.into())?;
+            } else {
+                let mut fixes = Vec::new();
+
+                let mut accum_line_num = AccumulateLineNum::new();
+                for typo in check_bytes(&bc, policy) {
+                    let line_num = accum_line_num.line_num(&buffer, typo.byte_offset);
+                    let (line, line_offset) = extract_line(&buffer, typo.byte_offset);
+                    let msg = report::Typo {
+                        context: Some(report::FileContext { path, line_num }.into()),
+                        buffer: std::borrow::Cow::Borrowed(line),
+                        byte_offset: line_offset,
+                        typo: typo.typo.as_ref(),
+                        corrections: typo.corrections.clone(),
+                    };
+                    // HACK: we use the reporter to display the possible corrections to the user
+                    // this will be looking very ugly with the format set to anything else than json
+                    // technically we should only report typos when not correcting
+                    reporter.report(msg.into())?;
+
+                    if let Some(correction_index) = select_fix(&typo) {
+                        fixes.push((typo, correction_index));
+                    }
+                }
+
+                if !fixes.is_empty() || path == std::path::Path::new("-") {
+                    let buffer = fix_buffer(buffer, fixes.into_iter());
+                    write_file(path, content_type, buffer, reporter)?;
+                }
+            }
+        }
+
+        if policy.check_filenames {
+            if let Some(file_name) = path.file_name().and_then(|s| s.to_str()) {
+                let mut fixes = Vec::new();
+
+                for typo in check_str(file_name, policy) {
+                    let msg = report::Typo {
+                        context: Some(report::PathContext { path }.into()),
+                        buffer: std::borrow::Cow::Borrowed(file_name.as_bytes()),
+                        byte_offset: typo.byte_offset,
+                        typo: typo.typo.as_ref(),
+                        corrections: typo.corrections.clone(),
+                    };
+                    // HACK: we use the reporter to display the possible corrections to the user
+                    // this will be looking very ugly with the format set to anything else than json
+                    // technically we should only report typos when not correcting
+                    reporter.report(msg.into())?;
+
+                    if let Some(correction_index) = select_fix(&typo) {
+                        fixes.push((typo, correction_index));
+                    }
+                }
+
                 if !fixes.is_empty() {
                     let new_path = fix_file_name(path, file_name, fixes.into_iter())?;
                     std::fs::rename(path, new_path)?;
@@ -673,6 +755,35 @@ fn fix_file_name<'a>(
     let new_name = String::from_utf8(new_name).expect("corrections are valid utf-8");
     let new_path = path.with_file_name(new_name);
     Ok(new_path)
+}
+
+fn select_fix(typo: &typos::Typo<'_>) -> Option<usize> {
+    let corrections = match &typo.corrections {
+        typos::Status::Corrections(c) => c,
+        _ => return None,
+    };
+
+    if corrections.len() == 1 {
+        Confirm::new()
+            .with_prompt("Do you want to apply the fix suggested above?")
+            .default(true)
+            .show_default(true)
+            .interact()
+            .ok()?;
+
+        Some(0)
+    } else {
+        let mut items = corrections.clone();
+
+        items.insert(0, std::borrow::Cow::from("None (skip)"));
+        let selection = Select::new()
+            .with_prompt("Please choose one of the following suggestions")
+            .items(&items)
+            .default(0)
+            .interact()
+            .ok()?;
+        selection.checked_sub(1)
+    }
 }
 
 pub fn walk_path(
--- a/crates/typos-cli/src/file.rs
+++ b/crates/typos-cli/src/file.rs
@@ -87,7 +87,7 @@ impl FileChecker for FixTypos {
                 let mut accum_line_num = AccumulateLineNum::new();
                 for typo in check_bytes(&buffer, policy) {
                     if is_fixable(&typo) {
-                        fixes.push(typo.into_owned());
+                        fixes.push((typo.into_owned(), 0));
                     } else {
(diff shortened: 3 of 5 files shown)


crates/typos-cli/src/bin/typos-cli/args.rs
```diff
@@ -67,6 +67,10 @@ pub(crate) struct Args {
     #[arg(long, short = 'w', group = "mode", help_heading = "Mode")]
     pub(crate) write_changes: bool,
 
+    /// Prompt for each suggested correction whether to write the fix
+    #[arg(long, short = 'a', group = "mode", help_heading = "Mode")]
+    pub(crate) write_ask: bool,
```
epage, line 72: Generally this would be `-i` `--interactive`

crates/typos-cli/src/file.rs
```diff
@@ -127,12 +129,85 @@ impl FileChecker for FixTypos {
...
+                }
+            }
+        }
+
+        Ok(())
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct AskFixTypos;
```
epage, line 142: Again, `Interactive`

crates/typos-cli/src/file.rs
```diff
@@ -127,12 +129,85 @@ impl FileChecker for FixTypos {
...
-                        String::from_utf8(new_name).expect("corrections are valid utf-8");
-                    let new_path = path.with_file_name(new_name);
-                    std::fs::rename(path, new_path)?;
+                    fix_file_name(path, &file_name.to_owned(), fixes)?;
+                }
+            }
+        }
+
+        Ok(())
+    }
```
epage, line 138: Please split refactors out into their own commits before the feature work

crates/typos-cli/src/file.rs
```diff
@@ -127,12 +129,85 @@ impl FileChecker for FixTypos {
...
+                    let msg = report::Typo {
+                        context: Some(report::PathContext { path }.into()),
+                        buffer: std::borrow::Cow::Borrowed(file_name.as_bytes()),
+                        byte_offset: typo.byte_offset,
+                        typo: typo.typo.as_ref(),
+                        corrections: typo.corrections.clone(),
+                    };
+                    reporter.report(msg.into())?;
+
+                    if let Some(correction_index) = select_fix(&typo) {
```
epage, line 202: We're only supposed to report things that aren't fixed

crates/typos-cli/src/file.rs
```diff
@@ -127,12 +129,85 @@ impl FileChecker for FixTypos {
...
+                        fixes.push((typo, correction_index));
+                    }
+
+                    println!("\n");
```
epage, line 206: Why this newline?

crates/typos-cli/src/file.rs
```diff
@@ -650,6 +725,40 @@ fn is_fixable(typo: &typos::Typo<'_>) -> bool {
...
+    file_name: &'a str,
+    fixes: Vec<(typos::Typo<'a>, usize)>,
+) -> Result<(), std::io::Error> {
+    let file_name = file_name.to_owned().into_bytes();
+    let new_name = fix_buffer_with_correction_index(file_name, fixes.into_iter());
+    let new_name = String::from_utf8(new_name).expect("corrections are valid utf-8");
+    let new_path = path.with_file_name(new_name);
+    std::fs::rename(path, new_path)?;
+    Ok(())
+}
```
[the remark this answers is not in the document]
epage: Feel free to do so.  Please generalize the function in a refactor commit before the feature work

crates/typos-cli/src/file.rs
```diff
@@ -664,6 +773,56 @@ fn fix_buffer(mut buffer: Vec<u8>, typos: impl Iterator<Item = typos::Typo<'stat
...
+    let file_name = file_name.to_owned().into_bytes();
+    let new_name = fix_buffer(file_name, fixes.into_iter());
+    let new_name = String::from_utf8(new_name).expect("corrections are valid utf-8");
+    let new_path = path.with_file_name(new_name);
+    std::fs::rename(path, new_path)?;
+    Ok(())
+}
+
+fn select_fix(typo: &typos::Typo<'_>) -> Option<usize> {
+    if is_fixable(&typo) {
```
epage, line 790: Instead of using `is_fixable`, extract the corrections and condition this on the length.

crates/typos-cli/src/file.rs
```diff
@@ -664,6 +773,56 @@ fn fix_buffer(mut buffer: Vec<u8>, typos: impl Iterator<Item = typos::Typo<'stat
...
+            .items(&items)
+            .default(0)
+            .interact()
+            .unwrap();
+
+        if selection == 0 {
+            return None;
+        }
+
+        return Some(selection - 1);
```
epage, line 820:
```suggestion
        .clone();

        items.insert(0, std::borrow::Cow::from("None (skip)"));
        let selection = Select::new()
            .with_prompt("Please choose one of the following suggestions")
            .items(&items)
            .default(0)
            .interact()
            .unwrap();
        if selection == 0 {
            return None;
        }
        return Some(selection - 1);
```

By grouping using blank lines, it helps clarify that these are all coupled together.  For example, I originally missed `items.insert` and was assuming `selection - 1` was a quirk of that API.

crates/typos-cli/src/file.rs
```diff
@@ -664,6 +773,56 @@ fn fix_buffer(mut buffer: Vec<u8>, typos: impl Iterator<Item = typos::Typo<'stat
...
+) -> Result<(), std::io::Error> {
+    let file_name = file_name.to_owned().into_bytes();
+    let new_name = fix_buffer(file_name, fixes.into_iter());
+    let new_name = String::from_utf8(new_name).expect("corrections are valid utf-8");
+    let new_path = path.with_file_name(new_name);
+    std::fs::rename(path, new_path)?;
+    Ok(())
+}
+
+fn select_fix(typo: &typos::Typo<'_>) -> Option<usize> {
```
epage, line 789: When I run this locally, I get weird output on the right item

image

crates/typos-cli/src/file.rs
Bnyro: To clearly have a visual separation between the prompt/selection about the current fix and the previous fix. Can be removed though if you fix it's unnecessary.

crates/typos-cli/src/file.rs
Bnyro: "not fixed" means typos that only have one possible correction? Shouldn't we prompt whether the user wants to apply that single possible fix?

Or do you mean running `reporter.report(msg.into())?`? That part however displays the suggested fix to the user, doesn't it?

Bnyro:
Thanks for your review so far, I'm way more comfortable with the code now than before the review.

crates/typos-cli/src/file.rs
```diff
@@ -650,10 +642,24 @@ fn is_fixable(typo: &typos::Typo<'_>) -> bool {
     extract_fix(typo).is_some()
 }
 
-fn fix_buffer(mut buffer: Vec<u8>, typos: impl Iterator<Item = typos::Typo<'static>>) -> Vec<u8> {
+fn fix_buffer(
+    mut buffer: Vec<u8>,
+    typos: Vec<typos::Typo<'_>>,
+    correction_indices: Option<Vec<usize>>,
```
epage, line 729: I feel like instead of parallel vec's, we should just have one iterator with the typo and index in a tuple.  For the existing cases, we could pass in `0`.

crates/typos-cli/src/file.rs
```diff
@@ -664,6 +670,19 @@ fn fix_buffer(mut buffer: Vec<u8>, typos: impl Iterator<Item = typos::Typo<'stat
     buffer
 }
 
+fn fix_file_name<'a>(
```
epage, line 754: Since you are already going back and making a change, could you split the extraction of `fix_file_name` out into its own commit before the adding of indices?  Each commit should do only one thing.  This is a big help for reviewers as I don't have to try to untangle the root reason for each change

crates/typos-cli/src/file.rs
epage: reporting is meant for use with `--format`.  It seems you are trying to take advantage of the output formatting for outputting the part of the prompt.  Technically, this will break down for json output but people probably shouldn't use this for progammatic output.

We should at least have a `// HACK:` comment explaining the abuse of reporting for the sake of prompting the user and that technically we should only report when not correcting.

crates/typos-cli/src/file.rs
```diff
@@ -683,6 +764,42 @@ fn fix_file_name<'a>(
...
+            .default(true)
+            .show_default(true)
+            .interact()
+            .unwrap();
+
+        if confirmation {
+            return Some(0);
+        }
+    } else {
+        let mut items = corrections.clone();
```
epage, line 786: nit: if we've cloned the corrections on 772, why do we need this clone?  We could just do `let mut items = corrections;`

crates/typos-cli/src/file.rs
Bnyro: > It seems you are trying to take advantage of the output formatting for outputting the part of the prompt. 
>
Oh well, I see, I didn't notice it's possible to specify other output format than the `long` format.

crates/typos-cli/src/file.rs
epage: Something still isn't right
image

And if I ctrl-c,. my cursor no longer shows up

Bnyro: I can reproduce the issue with the cursor (but thought something in my setup is broken, apparently it's not my setup), I've really no idea what causes it. I've tried the `inquire` library as well because I thought something is wrong with the `dialoguer` library but that one worked even worse...

Bnyro: Oh well, apparently we have to handle Ctrl+C issues ourselves as per https://github.com/console-rs/dialoguer/issues/294.

Bnyro: It seems like the available cli libraries for prompts are not really that mature yet.

Bnyro:
I've added a custom `ctrlc` handler now to handle manual program exits and added proper error handling to the `dialoguer` confirm and selection prompts.

I've no idea why the position of the prompt is weird sometimes.

crates/typos-cli/src/bin/typos-cli/main.rs
```diff
@@ -32,6 +32,13 @@ fn run() -> proc_exit::ExitResult {
 
     init_logging(args.verbose.log_level());
 
+    // https://github.com/console-rs/dialoguer/issues/294
+    ctrlc::set_handler(move || {
+        let _ = console::Term::stdout().show_cursor();
+        std::process::exit(0);
+    })
+    .expect("ctrlc::set_handler");
```
epage, line 40: an "expect" should say what assumption it is operating under for why its ok.

crates/typos-cli/src/bin/typos-cli/main.rs
```diff
@@ -32,6 +32,13 @@ fn run() -> proc_exit::ExitResult {
 
     init_logging(args.verbose.log_level());
 
+    // https://github.com/console-rs/dialoguer/issues/294
```
epage, line 35: It would help to put in a bit more detail, like
```rust
// HACK: Ensure the terminal gets reset to a good state if the user hits ctrl-c during a prompt
// see https://github.com/console-rs/dialoguer/issues/294 for more details
```

crates/typos-cli/src/file.rs
```diff
@@ -664,6 +745,53 @@ fn fix_buffer(mut buffer: Vec<u8>, typos: impl Iterator<Item = typos::Typo<'stat
...
+        _ => return None,
+    };
+
+    if corrections.len() == 1 {
+        let confirmation = Confirm::new()
+            .with_prompt("Do you want to apply the fix suggested above?")
+            .default(true)
+            .show_default(true)
+            .interact();
+        return confirmation.map(|_| 0).ok();
```
epage, line 772: imo use of `?` would make this easier to read

```suggestion
        let confirmation = Confirm::new()
            .with_prompt("Do you want to apply the fix suggested above?")
            .default(true)
            .show_default(true)
            .interact()
            .ok()?;
        0
```

crates/typos-cli/src/file.rs
```diff
@@ -664,6 +745,53 @@ fn fix_buffer(mut buffer: Vec<u8>, typos: impl Iterator<Item = typos::Typo<'stat
...
+        return match selection {
+            Ok(selection) => {
+                if selection != 0 {
+                    Some(selection - 1)
+                } else {
+                    None
+                }
+            }
+            Err(_) => None,
+        };
```
epage, line 791: Similar
```suggestion
            .interact()
            .ok();
        let index = selection.checked_sub(1)?;
        index
```

crates/typos-cli/src/bin/typos-cli/main.rs
```diff
@@ -32,6 +32,13 @@ fn run() -> proc_exit::ExitResult {
...
+    ctrlc::set_handler(move || {
+        let _ = console::Term::stdout().show_cursor();
```
epage, line 37: I now see the cursor but I can't see anything I type

crates/typos-cli/src/bin/typos-cli/main.rs
```diff
@@ -32,6 +32,13 @@ fn run() -> proc_exit::ExitResult {
...
+        std::process::exit(0);
```
epage, line 38: Is `exit(0)` the correct way of exiting on ctrl-c to mirror behavior before this change?

epage:
> I've no idea why the position of the prompt is weird sometimes.

For me, its always weird for the first prompt.  This would be a blocker for merging

Bnyro:
> For me, its always weird for the first prompt. This would be a blocker for merging
>
I see, but unfortunately I don't see any way to fix this except for trying other libraries?

epage:
How about we switch to showing prompts that require hitting `<enter>` like `git add -p` does?

Bnyro:
> How about we switch to showing prompts that require hitting `<enter>` like `git add -p` does?
>
That could fix the issue for the `confirm` prompt. If I run `typos --interactive` on this repository (crate-ci/typos), the first prompt is a selection prompt which is getting skipped immediately (due to the bug mentioned), although the `select` prompts should in theory require pressing `<enter>` as well. That's quite odd.

If I do however run this on different source code folders, for some everything works normally, and for some it behaves as described above, which doesn't make a lot of sense to me.

The only two considerable crates I found are the `dialoguer` crate (in use right now) and the `inquire` crate, which wouldn't even let me use the `Confirm` prompt properly, I wasn't able to actually enter yes or no when I tested it.

crates/typos-cli/src/bin/typos-cli/main.rs
Bnyro: That's strange, for me everything behaves normal after that change.

crates/typos-cli/src/bin/typos-cli/main.rs
Bnyro: You're right, the previous exit code for `ctrl-c` was 130, and according to my quick research 130 is preferred over 0 when `ctrl-c` killed the process via SIGTERM.

PR #1136, opened 2024-10-30
chore: Fix typo "potemtial" -> "potential"

PigeonF:


epage:
Always fun when these are found since I've not tried to exclude all of our test data in a way to actually allow typos to run against its own repo.

PR #1154, opened 2024-11-24
feat: Recognize dtso as devicetree type

dseight:
This adds dtso (device tree source overlay) as a recognized devicetree type.


epage:
Thanks for contributing!

This made me wonder what the `ignore` crate does and I realized we were several years out of date so I updated typos in #1155.  This added a `devicetree` type didn't add all of the extensions you specified.  So a rebase and this should be good to go!

dseight:
I looked yesterday and also noticed that it was a bit out of date. And my first thought was "okay, I'll just copy-paste the new file from ripgrep", but stumbled upon the format change first, and then also noticed that there were some changes to the original file... So, decided not to bother too much. 😁 Thank you for updating it!

PR #1169, opened 2024-12-07
fix: remove some unused deps

klensy:
This removes some unused/dev-only deps around.

Local tests passed, lets see.


klensy:
nano ping

PR #1173, opened 2024-12-12
Update README.md for better readability

Ghaniyyat05:
To add comments for better understanding of commands and updated header for a section


epage:
This is still doing two different things (adding explanations to code blocks and changing case)

PR #1116, opened 2024-10-08
feat(gha): add docker push github action

georgettica:
fixes #427

PR #1183, opened 2024-12-15
feat(infra): Publish Docker image on release

Alex-deVis:
The post-release job will trigger the publish-docker-image job which builds and pushes the image to the GitHub Container Registry (ghcr.io). The image is tagged with the `semver tag` and `latest`.

Regarding https://github.com/crate-ci/typos/issues/427#issuecomment-1031618262, even a focused image can be helpful for local development and testing. I think it would be beneficial to start publishing the image even if it might change in the future.

Fixes: #427

### DIFF
--- a/.github/workflows/publish-docker-image.yml
+++ b/.github/workflows/publish-docker-image.yml
@@ -1,0 +1,75 @@
+# The way this works is the following:
+#
+# The publish-docker-image job runs when a release is created, hence after post-release job.
+# The steps are:
+#  - Log in to the Container registry (example: ghcr.io/create-ci).
+#  - Extract metadata (tags, labels) for Docker image.
+#  - Build and push Docker image to the container registry.
+#    Image tags are "latest" and the semver tag (without the "v" prefix).
+#  - Generate artifact attestation.
+#
+# Reference:
+# https://docs.github.com/en/actions/use-cases-and-examples/publishing-packages/publishing-docker-images#publishing-images-to-github-packages
+
+name: publish-docker-image
+on:
+  workflow_run:
+    workflows: [post-release]
+    types:
+      - completed
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


epage:
At my quick glance, this seems very different than the workflow from #1116.  As the person purposing this alternative workflow, could you step through why you made different decisions?

Alex-deVis:
Hello, @epage, thanks for taking the time to review! My PR only addresses building the docker image on release to be able to run the action locally similarly to how it is run on GH actions pipeline. I have no interest in updating the Dockerfile. Therefore, you could say that the PRs are complementary; in my view, changes on Dockerfile should be separated from changes on actions. 

The backbone for my changes is this [guide from GitHub](https://docs.github.com/en/actions/use-cases-and-examples/publishing-packages/publishing-docker-images#publishing-images-to-github-packages). Let's dive into my changes:

- the trigger is the completion of `post-releaes` workflow. I think this is cleaner than `release: published` since creating a release consists only of creating the page, without any other validation. Bundling the completion of `post-releases` with the creation of the Docker image ensures that you'd only do it when everything is set and done.

- `Extract tag from master` step is needed because we use the `post-release` completion as the trigger. We have to determine the latest tag ourselves since the trigger is not tied to a tag or a PR. `git tag --merged ${{ github.ref_name }}` checks the latest tag on the master branch since that is where releases will be.

- `Extract metadata (tags, labels) for Docker` step adds the previous tag (`type=semver,pattern={{version}},value=${{ env.TAG }}`) and `latest` (`type=raw,value=latest,enable={{is_default_branch}}`) to the list of tags.

Alex-deVis:
Hi, @epage! Bump on this.

epage:
Please don't bump. I have a queue I'm working through, made worse from the holidays, and this creates extra noise.

epage:
That is very rude to re-bump a PR when you've already been told not to bump it.

nogweii:
A couple of comments from a scan:

* Is it necessary to put ghcr.io into an environment variable? I think it'd likely be the only registry that a typos image gets published to, but I may be wrong.
* This doesn't appear to build a multi-arch image, so users on arm64 platforms would miss out.

PR #1186, opened 2024-12-19
docs: add linkable header to example configurations

friday:
Config Fields is quite a section, and if you want to make a reference directly to the very helpful extend-ignore-re examples today you can't.


epage:
Thanks!

PR #1192, opened 2024-12-30
use choco to download wget

### LINKED ISSUE #1191: GH action not working for windows because of wget
This actions assumes that `wget` is already installed which is not true for `windows-latest`

This produces the error

```
D:\a\_actions\crate-ci\typos\v1.28.4/action/entrypoint.sh: line 43: wget: command not found
```

The action requires
```yaml
    - name: Install wget for Windows
      if: matrix.os == 'windows-latest'
      run: choco install wget --no-progress
```

Little bit related to #1095

Its-Just-Nans:
Fix #1191

Related to #1095

### DIFF
--- a/.github/workflows/test-action.yml
+++ b/.github/workflows/test-action.yml
@@ -12,10 +12,6 @@ jobs:
     - name: Checkout Actions Repository
       uses: actions/checkout@v4
 
-    - name: Install wget for Windows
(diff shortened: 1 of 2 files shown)


epage (changes requested):
Blocked on discussion in #1191

PR #1224, opened 2025-02-02
Add force_exclude to github action.

RichardBrown384:
Hi! I"m not sure if this is the right thing to do but I thought I'd open this to get some feedback.

The use case is to forcibly exclude files (like localisation files that are excluded in the configuration) that are modified by some automation process.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -25,6 +25,11 @@ inputs:
     required: false
     default: false
 
+  force_exclude:
+    description: "Respect excluded files when passed explicitly"
+    required: false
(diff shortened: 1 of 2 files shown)


epage:
As we note in our [contributing guide](https://github.com/crate-ci/typos/blob/master/CONTRIBUTING.md#pull-requests), we ask for features to be discussed in Issues (problem/solution focysed) before moving onto PRs (implementation focused).  This helps make sure we understand the problem and find the best-aligned solution.

Issue #1221, opened 2025-01-31
Non-critical word correction additions - February 2025 Edition


epage:
Please list any corrections you noticed `typos` doesn't fix and we'll roll these up into a single PR, making it lighter weight for adding new corrections.

Preferred format (for easy copy/pasting):
````
```
bad1 -> good1
bad2 -> good2
```
````
*(using a code block allows me one-click copy/paste)*

Feel free to update your comment as you find more suggestions for this month to reduce noise and making updating the dictionary easier

Please use dedicated issues for false positives as they likely need more discussion than is intended for this thread.

Related open issues
- #466
- #515
- #592
- #743
- #745
- #795

**Comment by szepeviktor** [comment]
```
descrtuctive -> destructive
searachable -> searchable
authentitaction -> authentication
incidates -> indicates
```

**Comment by cmdcolin** [comment]
```
raplidly -> rapidly
```

**Comment by tjkuson** [comment]
```
confirure -> configure
```

**Comment by cbachhuber** [comment]
```
integraions -> integrations
```

**Comment by diego-pm** [comment]
```
fligth -> flight
fligths -> flights
```

**Comment by AMS21** [comment]
```
folloed -> followed
```

**Comment by epage** [comment]
```
compatiblity → compatibility
stablized → stabilized
panicing → panicking
verbage → verbiage
corectly → correctly
```

**Comment by kachick** [comment]
```
delegrate -> delegate
```

**Comment by justanotheranonymoususer** [comment]
```
cnovert -> convert
```

**Comment by efaulhaber** [comment]
```
bidrectional -> bidirectional
```

**Comment by brody4hire** [comment]
```
exmapl -> example
ccatch -> catch
srry -> sorry
```

---

potentially controversial:

```
typs -> typos
typ -> typo
```

**Comment by aDotInTheVoid** [comment]
```
construstor -> constructor
```

**Comment by airwoodix** [comment]
```
partiy -> parity
```

**Comment by rparrett** [comment]
```
sypported -> supported
```

**Comment by lenaploetzke** [comment]
```
collelective -> collective
fre -> free
whenenever -> whenever
simarly -> similarly
ownded -> owned
```

**Comment by phip1611** [comment]
```
methology -> methodology
```

**Comment by DJMcNab** [comment]
```
earller -> earlier
```

**Comment by darkblaze69** [comment]
```
schaedule -> schedule
schaeduler -> scheduler
schaeduled -> scheduled

```

Crown0815:
[...]

PR #1231, opened 2025-02-13
Do not correct seeked word

dmtrKovalenko:
The word seeked is used for simplicity in the [web api](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/seeked_event) and when you work a lot with different video and media related project on the web it becomes really annoying that typos is always referring this in every project (and actually throws an error even when you open lib.dom.ts the typescript bindings to web api). 

I tried to check your misspelled implementation but it looks it using some totally abandoned project https://github.com/client9/misspell so I decided to just exclude it in the most naive way. Please let me know if something lese needs to be done

### DIFF
--- a/crates/wikipedia-dict/assets/dictionary.txt
+++ b/crates/wikipedia-dict/assets/dictionary.txt
@@ -3477,7 +3477,6 @@ seceeded->succeeded, seceded
 secratary->secretary
 secretery->secretary
 sedereal->sidereal
-seeked->sought
 segementation->segmentation
 seguoys->segues
 seige->siege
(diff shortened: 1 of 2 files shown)


crates/wikipedia-dict/assets/dictionary.txt
```diff
@@ -3477,7 +3477,6 @@ seceeded->succeeded, seceded
 secratary->secretary
 secretery->secretary
 sedereal->sidereal
-seeked->sought
```
epage, line 3480: This should be reverted.  We only use `typos-dict`.  The others are for comparison purposes

epage:
Per the instructions in #1221 and our [contributing guide](https://github.com/crate-ci/typos/blob/master/CONTRIBUTING.md#pull-requests), could you open an issue first?  We focus Issues on getting consesus on problems and solutions and PRs for implementation.

PR #1250, opened 2025-03-02
chore(ci): maintain major tags for GitHub Actions

### LINKED ISSUE #857: Maintain `v1` tag for Github Actions?
It would be great to have a `v1` tag so that actions automatically takes advantage of upstream non-semver breaking changes without having to bump the action frequently. That is all 😄

muzimuzhi:
Fixes #857.

### DIFF
--- a/.github/workflows/post-release.yml
+++ b/.github/workflows/post-release.yml
@@ -138,3 +138,7 @@ jobs:
       env:
         GH_TOKEN: ${{ github.token }}
       run: gh release edit "${{ needs.create-release.outputs.tag }}" --draft=false
+    - name: Update major tag
+      uses: rickstaa/action-update-semver@v1
+      with:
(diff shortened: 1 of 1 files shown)


.github/workflows/post-release.yml
```diff
@@ -125,6 +125,10 @@ jobs:
       run: |
         tag="${{ needs.create-release.outputs.tag }}"
         gh release upload "$tag" ${{ env.ASSET }}
+    - name: Update major tag
+      uses: rickstaa/action-update-semver@v1
+      with:
+        major_version_tag_only: true
```
epage, line 131: This needs to be in the `publish-release` job

muzimuzhi: Moved it to the end of the `publish-release` job.

PR #1261, opened 2025-03-24
fix(type): Include support for go.work

### LINKED ISSUE #1260: False positive for go.work.sum
### Description

```
error: `Ue` should be `Use`, `Due`
  --> ./utils/go.work.sum:813:55
    |
813 | github.com/nxadm/tail v1.4.8/go.mod h1:+ncqLTQzXmGhMZNUePPaPqPvBxHAIsmXswZKocGu+AU=
    |                                                       ^^
    |
```

I believe [this](https://github.com/crate-ci/typos/blob/8c7c9e5c7c0bcdb57c3452e29f60ccf3a34b66e2/crates/typos-cli/src/default_types.rs#L168) should be extended to include `go.work.sum`. To support [Go workspaces](https://go.dev/ref/mod#workspaces) in general there also must be something that treats `go.work` as `go.mod` and applies the same rules to it, I couldn't find anything related to `go.mod` in the above file for some reason.

epage:
Fixes #1260


azahi:
@epage This solves my issue. Thanks for a quick fix!

PR #1274, opened 2025-04-08
docs(readme): update the toc

kemingy:


epage:
Could you help me understand why you feel this new toc is needed?

kemingy:
> Could you help me understand why you feel this new toc is needed?

The old toc doesn't match the current README content.

However, since the current README is short and GitHub already has a built-in TOC, I could change the PR to remove the TOC if you like.

epage:
Note that this is under "Documentation".  This is not meant to be a ToC of the current `README.md` but a ToC of all of our documentation.

kemingy:
> Note that this is under "Documentation". This is not meant to be a ToC of the current `README.md` but a ToC of all of our documentation.

Got it.

PR #1296, opened 2025-05-12
docs: fix typos

omahs:
docs: fix typos


epage:
This is a vendored document and we shouldn't edit it.

PR #1328, opened 2025-06-19
fix(cli): Don't recognize "typ" as typo in .typ files

### LINKED ISSUE #1253: `typ` as Typst extension
### Description

`typ -> typo` from https://github.com/crate-ci/typos/issues/1221#issuecomment-2648925186 should be reverted since `typ` is the extension of Typst files.

`typ` could also be `type`, so the current state is problematic anyway.

not-my-profile:
Before the file extension of Typst was recognized as a typo.

Fixes #1253.

### DIFF
--- a/crates/typos-cli/src/default_types.rs
+++ b/crates/typos-cli/src/default_types.rs
@@ -315,6 +315,7 @@ pub(crate) const DEFAULT_TYPES: &[(&str, &[&str])] = &[
     ("ts", &["*.ts", "*.tsx", "*.cts", "*.mts"]),
     ("twig", &["*.twig"]),
     ("txt", &["*.txt"]),
+    ("typ", &["*.typ"]),
     ("typoscript", &["*.typoscript"]),
     ("usd", &["*.usd", "*.usda", "*.usdc"]),
     ("v", &["*.v", "*.vsh"]),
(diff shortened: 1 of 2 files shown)


crates/typos-cli/src/default_types.rs
```diff
@@ -315,6 +315,7 @@ pub(crate) const DEFAULT_TYPES: &[(&str, &[&str])] = &[
     ("ts", &["*.ts", "*.tsx", "*.cts", "*.mts"]),
     ("twig", &["*.twig"]),
     ("txt", &["*.txt"]),
+    ("typ", &["*.typ"]),
```
epage, line 318: This file is copied out of ripgrep
https://github.com/BurntSushi/ripgrep/blob/master/crates/ignore/src/default_types.rs

Changes to it could easily b lost when we do updates from ripgrep

lafrenierejm: It looks like this will make it into ripgrep in some time. It's currently staged in BurntSushi's `ag/rollup` branch following <https://github.com/BurntSushi/ripgrep/pull/2914>: <https://github.com/BurntSushi/ripgrep/blob/ag/rollup/crates/ignore/src/default_types.rs#L302>

epage: #1396 updates the type list from ripgrep

Issue #1363, opened 2025-08-25
Typos claims a typo in a version number in a go.mod file


neilschark:
### Please complete the following tasks

- [x] I have searched the [open](https://github.com/crate-ci/typos/issues?q=is%3Aissue%20state%3Aopen%20label%3AA-exclude) and [rejected](https://github.com/crate-ci/typos/issues?q=is%3Aissue%20state%3Aclosed%20label%3AA-exclude) issues

### Example data

github.com/dprotaso/go-yit v0.0.0-20220510233725-9ba8df137936 // indirect

### Incorrect correction

error: `ba` should be `by`, `be`

### Description of data format

A normal go.mod file for dependency management in go.

Example:

```
module test/test

go 1.24

require (
	github.com/dprotaso/go-yit v0.0.0-20220510233725-9ba8df137936
)
```

### Justification

A go.mod file is always used in golang.

I will just exclude the file from being checked, as there is not much use to it anyways.

### Notes

Output from our CI
```
$ typos
error: `ba` should be `by`, `be`
  --> ./go.mod:32:52
   |
32 |  github.com/dprotaso/go-yit v0.0.0-20220510233725-9ba8df137936 // indirect
   |                                                    ^^
   |
```

epage:
I'm assuming the hash `9ba8df137936` is below our threshold for being ignored

Our main issue for things like that are
- #415
- #484

A workaround is defining a `extend-ignore-re` to match that line.

The question from there is if we should have special casing to detect the go case. 

Note that for lockfiles
https://github.com/crate-ci/typos/blob/9c80e6305930cd242bdbd44fcf89588cf781865e/crates/typos-cli/src/default_types.rs#L163-L170
we exclude them from checking
https://github.com/crate-ci/typos/blob/9c80e6305930cd242bdbd44fcf89588cf781865e/crates/typos-cli/src/file_type_specifics.rs#L7

Its looking like `go.mod` isn't quite a lockfile but isn't quite a human edited file.

epage:
Huh, we had #683  which was fixed in #684 which doesn't quite look correct.  It just dropped `go.mod` from being considered a go file type but it didn't add it to any exclusion lists.

ccoVeille:
For record, the following files have a similar format

- go.mod
- go.sum
- go.work
- go.work.sum

If you want to work on that @epage

epage:
Do you have a primary source I can refer back to on that?

ccoVeille:
[...]

PR #684, opened 2023-03-13
fix: ignore go.mod by default

fuyufjh:
Fix #683

Issue #683, opened 2023-03-13
ignore `go.mod` by default 


fuyufjh:
#458 fixed the issue of `go.sum`, but actually `go.mod` has a similar problem. It saves indirect dependencies in the second `require (...)` block, which were generated by `go add` automatically. 

Example:

```go
require (
	github.com/99designs/keyring v1.1.6 // indirect
	github.com/AthenZ/athenz v1.10.39 // indirect
	github.com/DataDog/zstd v1.5.0 // indirect
	github.com/andres-erbsen/clock v0.0.0-20160526145045-9e14626cd129 // indirect
	github.com/apache/pulsar-client-go/oauth2 v0.0.0-20220120090717-25e59572242e // indirect
	github.com/ardielle/ardielle-go v1.5.2 // indirect
	github.com/beorn7/perks v1.0.1 // indirect
	github.com/cespare/xxhash/v2 v2.1.2 // indirect
	...
```

Today I ran into this problem:

```
Run crate-ci/typos@v1.13.20
...
$ typos .
Warning: "afe" should be "safe".
error: `afe` should be `safe`
  --> ./integration_tests/datagen/go.mod:40:61
   |
40 |  github.com/gsterjov/go-libsecret v0.0.0-20161001094[7](https://github.com/risingwavelabs/risingwave/actions/runs/4403015548/jobs/7710853514#step:4:8)33-a6f4afe4[9](https://github.com/risingwavelabs/risingwave/actions/runs/4403015548/jobs/7710853514#step:4:10)[10](https://github.com/risingwavelabs/risingwave/actions/runs/4403015548/jobs/7710853514#step:4:11)c // indirect
   |                                                             ^^^
   |
```

https://github.com/risingwavelabs/risingwave/actions/runs/4403015548/jobs/7710853514

epage:
Unsure why I didn't catch this until #1363 but this didn't cause us to ignore `go.mod` but just not categorize it as Go code, so #1365 is reverting this.

#1363 is discussing the hashes in `go.mod`

PR #1365, opened 2025-08-28
fix: Revert "fix: ignore go.mod by default"

epage:
This reverts commit 4eeb460bb7575dcee03f2419e7331f71bc27b7c5.

Original PR: #684

Original Issue: #683

This didn't cause `go.mod` to be ignored instead it caused it to not be counted as go code.
However, `go.mod` and `go.work` are similar file types and both are more of package specifications than lock files or other machine-only content.

See #1363

### DIFF
--- a/crates/typos-cli/src/default_types.rs
+++ b/crates/typos-cli/src/default_types.rs
@@ -102,7 +102,7 @@ pub(crate) const DEFAULT_TYPES: &[(&str, &[&str])] = &[
     ("fut", &["*.fut"]),
     ("gap", &["*.g", "*.gap", "*.gi", "*.gd", "*.tst"]),
     ("gn", &["*.gn", "*.gni"]),
-    ("go", &["*.go", "go.work"]),
+    ("go", &["*.go", "go.mod", "go.work"]),
     ("gprbuild", &["*.gpr"]),
     ("gradle", &[
         "*.gradle", "*.gradle.kts", "gradle.properties", "gradle-wrapper.*",


ccoVeille:
Nitpicking: I would fix the typo in commit message 

`ignoredinstead` should be `ignored instead`

ccoVeille:
Is there a need to add `go.sum` and `go.work.sum` to ignored files ?

epage:
> Is there a need to add go.sum and go.work.sum to ignored files ?

Those are in the lockfile type which is not checked

ccoVeille:
This change equivalent to add locally add go.mod and go.work to extends glob for type go, right?

Because when I try it locally it reports typos for go.mod files

Try this one

https://github.com/noborus/ov/blob/master/go.mod

Maybe I'm wrong, but I prefer to ask.

epage:
Yes, this makes `*.go`, `go.mod`, and `go.work` all the same file type.  Unless you have some go-specific config in place, this shouldn't really change much.

ccoVeille:
> > Is there a need to add go.sum and go.work.sum to ignored files ?
> 
> Those are in the lockfile type which is not checked

Indeed. I didn't see it.

> Yes, this makes `*.go`, `go.mod`, and `go.work` all the same file type.  Unless you have some go-specific config in place, this shouldn't really change much.

Then, it reports errors.

I feel like in .go files everything is checked, and not only comments.

But we expect something different for go.mod and go.work where we only need to check comment.

Please try with the go.mod file I provided

epage:
> I feel like in .go files everything is checked, and not only comments.

`typos` works off of all file contents and has no knowledge of individual file types grammars, including comments.

ccoVeille:
> > I feel like in .go files everything is checked, and not only comments.
> 
> `typos` works off of all file contents and has no knowledge of individual file types grammars, including comments.

Oh OK. Then I would exclude go.mod and go.work totally.

That's what I did for many projects.

ccoVeille:
I just saw your comment here.

> this just becomes a hash detection issue.
_Originally posted by @epage in https://github.com/crate-ci/typos/issues/1363#issuecomment-3234313476_

You are right. Let's continue the discussion in #1363

- https://github.com/crate-ci/typos/issues/1363#issuecomment-3235882682

PR #1394, opened 2025-10-06
fix(dict): don't correct typ in Go files

ccoVeille:
type is a reserved word in Go.

As a consequence, typ is often used instead when handling types.

Here is a GitHub search about it [`language:go /[ \.]typ\W/`](https://github.com/search?q=language%3Ago%20%2F%5B%20%5C.%5Dtyp%5CW%2F&type=code)

Also, this term is often whitelisted on project using `typos`

[`language:toml path:/typos.toml/ /"typ"/`](https://github.com/search?q=language%3Atoml%20path%3A%2Ftypos.toml%2F%20%2F%22typ%22%2F&type=code)

Maybe more language could exclude `typ`, but I limited my PR to Go for now

### DIFF
--- a/crates/typos-cli/src/file_type_specifics.rs
+++ b/crates/typos-cli/src/file_type_specifics.rs
@@ -39,6 +39,7 @@ pub(crate) const TYPE_SPECIFIC_DICTS: &[(&str, StaticDictConfig)] = &[
         StaticDictConfig {
             ignore_idents: &[
                 "flate", // https://pkg.go.dev/compress/flate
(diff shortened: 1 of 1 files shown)


epage:
This is actually a larger conversation than just go, see #1253

ccoVeille:
Does it mean it cannot be addressed for Go ?

ccoVeille:
I commented https://github.com/crate-ci/typos/issues/1253#issuecomment-3373766459

Maybe conversation should continue there

ccoVeille:
for record, `typ` was removed from words to fix via

- #1397

PR #1403, opened 2025-10-21
feat: support for files which extend ignored words and/or identifiers

### LINKED ISSUE #931: Support plain text .dic dictionary files
Many projects like Chromium use standard `.dic` files to list all "known" words, i.e. those words that should NOT be corrected. Is it possible to add support for this?  Or is this something already supported (I couldn't find it in the readme or code search)

A .dic file is a simple text file with one word per line.  I don't recall how capitalization is specified (i.e. must be exact, or it allows a lower-cased word in the .dic file to be in upper-case to be ignored, but not the other way around).

Rolv-Apneseth:
This is an initial, very basic implementation for a solution that can hopefully close #931

Basically we just parse the given file(s) for either identifiers or words, simply broken up by _any_ white space, and use that to extend the words/identifiers, which is how I saw it done for the ignore words/identifiers from the file-specific configurations.

Currently there's no logic for considering something an invalid entry (e.g. multiple "words" per line), which I know was discussed in that issue but I'm not sure is necessary.

Just let me know if you want me to take a different approach, or if it needs refactoring etc. and I'll gladly make any changes. First time really working with this code so I'm not too familiar with it.

### DIFF
--- a/crates/typos-cli/src/bin/typos-cli/args.rs
+++ b/crates/typos-cli/src/bin/typos-cli/args.rs
@@ -54,6 +54,20 @@ pub(crate) struct Args {
     #[arg(short = 'c', long = "config", help_heading = "Config")]
     pub(crate) custom_config: Option<std::path::PathBuf>,
 
+    /// Path to a file specifying identifiers which are to be ignored .
+    ///
+    /// The specified file should be UTF-8 encoded, and identifiers should
+    /// be separated by whitespace.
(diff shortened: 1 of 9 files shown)


epage:
Converting to a Draft as there isn't anything decided in #931 yet.

Rolv-Apneseth:
Yep my bad meant to make it a draft

Issue #1175, opened 2024-12-13
Support more report formats


braineo:
Hello, thanks for the awesome work I think this tool saves my days, many days.

Currently in order to make typos integrate well with the source control system like gitlab, one needs to fiddle the json output like.

```bash
# this is my snippet for gitlab
typos --format json | jq --slurp '[.[] | select(.type == "typo") |
            {"description": "Typo detected: \(.typo) -> should be \(.corrections | join(", "))",
             "fingerprint": (.path + ":" + (.byte_offset | tostring)),
             "severity": "minor",
             "check_name": "typo",
             "location": {
             "path": .path,
               "lines": {
                 "begin": (.line_num // 1)
               }
              }
            }]' >typos-report.json
```

And I saw a ticket created two years ago for similar request

https://github.com/crate-ci/typos/issues/594

Take another tool I've been using for example, ruff, can generate different formats for the pipelines

https://docs.astral.sh/ruff/settings/#output-format

I think it could be beneficial to create a extensive architecture for contributors to implement different report formats?

epage:
As you said, we have #594 for Sarif which someone has started work on in #1047 which is in my backlog to get back to.

Someone also expressed interest for Code Climate in #937.

Looking at Ruff
- junit: not even sure what lints in junit would look like (I guess [this](https://github.com/astral-sh/ruff/blob/3533d7f5b4617ae91bc5cc3c4d0bbef4677be431/crates/ruff_linter/src/message/junit.rs#L15)) and I've had bad experience with how under-defined junit is, with it depending a lot on the tool reading it
- github, gitlab, azure:
  - github would also be covered by Sarif support
  - When discussing specific vendors, this gets into the question of what vendors do we support and how stable is the support within those vendors.

Huh, in the source they also support
- Sarif
- rdjson (reviewdog) ... something like trunk.io (#1037)?

The core of your issue is 

> I think it could be beneficial to create a extensive architecture for contributors to implement different report formats?

We do have some flexibility and #1047 highlights a gap we have (non-incremental output) and fills it.

So if that is the concern, then that is being taken care of.

The title is about supporting "more report formats".  That is overly broad for an issue.  If you have requests for specific formats, like gitlab, I would recommend being specific.

As there isn't anything more to act on within this issue, I'm going to go ahead and close it.

PR #1426, opened 2025-11-17
Feat(CLI): GitHub Actions Output

danjones1618:
This implements GitHub output directly in the CLI tool.
This also enables file name typos to be visualised.


epage:
Note that our [contrib docs](https://github.com/crate-ci/typos/blob/master/CONTRIBUTING.md#pull-requests) suggest proposing new features in issues first before implementation and posting of PRs.

In particular, this has been previously discussed at #429 and #1175.  Overall, we want to shift our github action to instead focus on SARIF reporting.

As this has been previously rejected, this is definitely not ready for the PR stage and I'm going to go ahead and close this PR.  If there is a reason for us to re-evaluate, that should be done in an Issue instead.

PR #1450, opened 2025-12-26
chore: Update attributes

### LINKED ISSUE #1449: improve PR with -152k / +152k changes
### Description

Considering PR where a lead to rebuild the wordlist

The file crates/typos-dict/src/word_codegen.rs would be updated in the PR

example: #1448

Is there a need to review this file in the PR ?

This file has a "this file is generated" header, but I wonder if it works, as it reports "large file are not rendered" instead of "generated file are not rendered"

Did you try using this in `.gitattributes` ?

```gitattributes
crates/typos-dict/src/word_codegen.rs linguist-generated=true
```

Or maybe this 

```gitattributes
crates/typos-dict/src/word_codegen.rs binary
```

epage:
Fixes #1449

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,4 @@
 # force LF on checkout to avoid rustfmt messing up the EOL
 crates/** text eol=lf
+crates/typos-*/src/*.rs linguist-generated=true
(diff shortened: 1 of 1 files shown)


ccoVeille:
Thanks

PR #1467, opened 2026-01-14
Use concrete examples in reference.md

Wilfred:
The examples for extend-ignore-re didn't show the full TOML syntax, and didn't always quote the string consistently. Show full examples that users can copy-paste into their `typos.toml`.


epage:
Thanks!

PR #1470, opened 2026-01-18
Upgrade pre-commit hooks

cclauss:
* https://github.com/pre-commit/pre-commit-hooks/releases
* https://github.com/crate-ci/committed/releases


epage:
I've posted epage/_rust#41 to update the hooks in our template which will eventually propagate to this repo.

Is there a reason you were interested in us updating our hooks such we should consider doing this now rather than waiting on the template update?

cclauss:
Eventual propagation works for me, given that pre-commit passes in both cases.

PR #1523, opened 2026-03-12
feat(infra): Publish Docker images on release

andrey-fomin:
Provide an official Docker Hub distribution for releases so users can pull typos without building from source or relying on stale third-party images.

Refs: #427
Related: #1116, #1183

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
     permissions:
       contents: none
     name: CI
-    needs: [test, wasm, msrv, lockfile, docs, rustfmt, clippy]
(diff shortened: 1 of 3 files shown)


andrey-fomin:
Before enabling this workflow, repo owners will need to:
- create a Docker Hub access token with permission to push to that repository
- add `DOCKERHUB_USERNAME` as a GitHub Actions variable
- add `DOCKERHUB_TOKEN` as a GitHub Actions secret

One thing to confirm before merging:
- what is the correct Docker Hub user/org account for the official image?
  I used `crateci/typos` as a placeholder since Docker Hub account names cannot contain.

For context, I also created `andrey-fomin/typos-docker` to build Docker images for typos:
https://github.com/andrey-fomin/typos-docker

PR #1539, opened 2026-04-13
fix(action): Leave binary in temp dir

### LINKED ISSUE #1534: Typos Action leaves a side effect binary post execution
### Description

In a job with multiple steps of which one is Typos, having a side effect (i.e. Typos binary being left behind post completion without cleanup), doesn't play well with succeeding steps (eg. REUSE). Could the default behavior be changed OR an input added to cleanup binary post execution?

epage:
Fixes #1534


avishj:
Thank you @epage!

PR #1542, opened 2026-04-27
docs(action): Update actions/checkout to v6 in examples

alexandear:
See https://github.com/actions/checkout/releases


epage:
Not thrilled with fielding PRs to update this every time it is changed.  People already need something on their side to help them ensure they are using new versions.  This also has several flaws (using a mutable tag, not pinned with a hash, persisting credentials).

If we update this, i would lean towards updating it right which may require setting up RenovateBot to update this.

PR #1550, opened 2026-05-01
fix(dict): Additional suggestions for `croped`, `sice`, `tendence`

DimitriPapadopoulos:
As suggested in https://github.com/crate-ci/typos/issues/1531#issuecomment-4353905571.

### DIFF
--- a/crates/typos-dict/assets/words.csv
+++ b/crates/typos-dict/assets/words.csv
@@ -24073,7 +24073,7 @@ croozed,cruised
 croozer,cruiser
 croozes,cruises
 croozing,cruising
-croped,coped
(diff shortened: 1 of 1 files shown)


epage:
I think there was a misunderstanding. I didn't say everything you had posted was not for that issue. I asked you to use the template. I said if it doesn't fit, then it *might* not be appropriate.

Those issues are fully additive. It is in removing items that it suggestions are no longer appropriate. In May's, I had already clarified in the top post that "this will be applied additively to any existing entries"

epage:
As this is incomplete and can be done through the issue, I'm going to go ahead and close this.

PR #1560, opened 2026-05-25
docs: better link for pr code suggestions tool

maartenh:
The previous link for submitting code suggestions linked to an old, unmaintained and archived action. The updated link goes to a better maintained action that performs the same work.


epage:
As this is migrating from Sentry to someone I'm not familiar with, I'd rather just remove the suggestion entirely than risk users using an actuion I can't vouch for.

Issue #1535, opened 2026-04-10
typos-cli Crash Report


eksperimental:
### Description

Looks like typos was trying to run on a temporary file that got deleted too quick or got deleted while typos was running and typos went crazy

[report-658f4a2a-24d4-4c5a-97d8-dd6fd3fa69a9.toml.txt](https://github.com/user-attachments/files/26620840/report-658f4a2a-24d4-4c5a-97d8-dd6fd3fa69a9.toml.txt)
[report-cfdcb386-8ca9-4c6a-bb83-3231f4c6f2c2.toml.txt](https://github.com/user-attachments/files/26620839/report-cfdcb386-8ca9-4c6a-bb83-3231f4c6f2c2.toml.txt)

```console
[error] ./coveralls.json.tmp: No such file or directory (os error 2)
Well, this is embarrassing.

typos-cli had a problem and crashed. To help us diagnose the problem you can send us a crash report.

We have generated a report file at "/var/folders/cg/.../T/report-cfdcb386-8ca9-4c6a-bb83-3231f4c6f2c2.toml". Submit an issue or email with the subject of "typos-cli Crash Report" and include the report as an attachment.


We take privacy seriously, and do not perform any automated error collection. In order to improve the software, we rely on people to submit reports.

Thank you kindly!
Well, this is embarrassing.

typos-cli had a problem and crashed. To help us diagnose the problem you can send us a crash report.

We have generated a report file at "/var/folders/cg/.../T/report-658f4a2a-24d4-4c5a-97d8-dd6fd3fa69a9.toml". Submit an issue or email with the subject of "typos-cli Crash Report" and include the report as an attachment.


We take privacy seriously, and do not perform any automated error collection. In order to improve the software, we rely on people to submit reports.

Thank you kindly!

[error] ./coveralls.json.tmp: No such file or directory (os error 2)
Well, this is embarrassing.

typos-cli had a problem and crashed. To help us diagnose the problem you can send us a crash report.

We have generated a report file at "/var/folders/cg/.../T/report-cfdcb386-8ca9-4c6a-bb83-3231f4c6f2c2.toml". Submit an issue or email with the subject of "typos-cli Crash Report" and include the report as an attachment.


We take privacy seriously, and do not perform any automated error collection. In order to improve the software, we rely on people to submit reports.

Thank you kindly!
Well, this is embarrassing.

typos-cli had a problem and crashed. To help us diagnose the problem you can send us a crash report.

[...]

Issue #1444, opened 2025-12-15
Panic: walk() should be called first


niebayes:
### Description

The detailed report is the following. No reproduction since I have not reproduced it yet.

```
typo-cli panic

name = "typos-cli"
operating_system = "Mac OS 26.0.1 [64-bit]"
crate_version = "1.38.1"
explanation = """
Panic occurred in file '/Users/sunflower/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/typos-cli-1.38.1/src/policy.rs' at line 95
"""
cause = "`walk()` should be called first"
method = "Panic"
backtrace = """
   0:        0x105003780 - backtrace::backtrace::trace::h53e956ce854338e4
   1:        0x10500c204 - backtrace::capture::Backtrace::create::h7df00c8ffa4f144a
   2:        0x10500c17c - backtrace::capture::Backtrace::new::hfed8a00fa5e13c59
   3:        0x104fd93a8 - human_panic::report::Report::new::hcb6e7f74a438913c
   4:        0x104fdb348 - human_panic::handle_dump::h64bc1c34504df8e7
   5:        0x104fb66fc - human_panic::setup_panic::{{closure}}::h913d7729d4e99850
   6:        0x10520e190 - std::panicking::rust_panic_with_hook::h557a23724a5de839
   7:        0x10520de38 - std::panicking::begin_panic_handler::{{closure}}::h269cace6208fef05
   8:        0x10520c68c - std::sys::backtrace::__rust_end_short_backtrace::h5be0da278f3aaec7
   9:        0x10520db14 - __rustc[de2ca18b4c54d5b8]::rust_begin_unwind
  10:        0x105247350 - core::panicking::panic_fmt::h477ff48eff31ffa4
  11:        0x105247330 - core::option::expect_failed::hb9685a494c775073
  12:        0x105073710 - typos_cli::policy::ConfigEngine::policy::h8fc3a88cbe8d744f
  13:        0x105047d30 - typos_cli::file::walk_entry::h43c68b06f1df4a4d
  14:        0x10506ad18 - typos_cli::file::walk_path_parallel::{{closure}}::{{closure}}::h46bd6232cf2b54db
  15:        0x1050a87e0 - <ignore::walk::FnVisitorImp as ignore::walk::ParallelVisitor>::visit::haf651c6661ffe93f
  16:        0x1050a9bb0 - ignore::walk::Worker::run::h06e220e306a747b6
[...]

PR #1577, opened 2026-07-01
fix(cli): don't panic when canonicalize() fails mid-walk

chatman-media:
Ran into #1444 while poking around the walker code. `walk_entry` canonicalizes every entry's path to look up its policy, and if that fails the error gets reported through `report_result`, which then falls back to `T::default()` — for a `PathBuf` that's an empty path. That empty path isn't absolute and never matches anything in `ConfigEngine`'s directory map, so `engine.policy()` panics with `` `walk()` should be called first `` right after we just told the user about a permission/IO error.

The easiest way to trigger `canonicalize()` failing on an entry the walker still hands back as `Ok` is a file disappearing between being listed and being canonicalized — same shape as the temp-file race in #1535, and probably what's behind the WAL-file crash report too (a permission-denied directory gets caught earlier by `ignore` itself, before it ever reaches this code).

Fix just skips the entry after reporting the error instead of continuing with garbage, same as the pattern already used a couple lines up for walk errors. Added a test that deletes a file right as the walker visits it and asserts we get a clean error instead of a panic — confirmed it panics on master and passes with the fix.

### DIFF
--- a/crates/typos-cli/tests/walk_permission_denied.rs
+++ b/crates/typos-cli/tests/walk_permission_denied.rs
@@ -1,0 +1,75 @@
+//! Regression test for <https://github.com/crate-ci/typos/issues/1444>
+//!
+//! `walk_entry` calls `Path::canonicalize()` on every walked entry to look up its policy.
+//! If a file disappears between the `ignore` walker listing it and `canonicalize()` running
+//! on it (e.g. a temp file removed mid-scan, as in issue #1535) -- or more generally
+//! whenever `canonicalize()` fails for a reason the walker itself didn't already catch --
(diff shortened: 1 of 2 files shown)


crates/typos-cli/tests/walk_permission_denied.rs
epage, line 1: To help demonstrate the problem, it would help if you added the test in the previous commit, reproducing the problem with it passing (likely a `should_panic = ""` on the test), see https://epage.github.io/dev/pr-style/#c-split for more details

epage:
Thanks for digging into this!

Issue #593, opened 2022-10-13
excludes are relative to current directory, not `typos.toml` location


embe-pw:
When `$DIR/typos.toml` contains 
```toml
[files]
extend-exclude = ["a/*.txt"]
```
and we run `typos` in `$DIR`, the file `a/test.txt` is correctly excluded. But when `typos` is run in `$DIR/a`, the exclude is not respected:
```console
$ cat typos.toml
[files]
extend-exclude = ["a/*.txt"]
$ typos
$ (cd a; typos)
error: `tpyo` should be `typo`
  --> ./test.txt:1:1
  |
1 | tpyo
  | ^^^^
  |
```

When `typos.toml` excludes `*.txt` both cases work, so the issue is not caused by ignoring the configuration in the parent directory:
```console
$ cat typos.toml
[files]
extend-exclude = ["*.txt"]
$ typos
$ (cd a; typos)
```

It would make sense to interpret the excludes relative to the `typos.toml` location.

(tested with `typos` 1.12.8 binary from the GitHub release)

epage:
If resolving this, a question would be what to do for `--config`.  See #861 where using the parent directory of the config wouldn't be correct.  Maybe we use the parent directory of the auto-discovered config only and make `--config` relative to that?

brandonkal:
It seems the toml file it is ignored if I call the file directly.
I find that typos often has false positives with SOPS files, for example within the BEGIN AGE ENCRYPTED FILE block.
Ideally, typos would recognize and ignore sops files, as most content is encrypted. That is easy to detect.
But for now I created a typos.toml in the root of the repository.
```toml
[files]
extend-exclude = ["*.sops.*"]
```
If I run `typos` I get no output but if I call the file directly:
```
❯ typos nomad/secrets.sops.yaml 
error: `Ba` should be `By`, `Be`
  --> nomad/secrets.sops.yaml:18:24
   |
18 |         NndQOEpqVWNyK1JBa2xTbjlHRmdXalpEVlhvCmVpNTdDRWZlN1ZRcUZmSkV6SjYy
   |                        ^^
   |
```
Which is unexpected because the config file exists. My lefthook pre-commit has `typos {staged_files}`

brandonkal:
Even calling `typos --config typos.toml nomad/secrets.sops.yaml` returns the same error. Which is its own bug. So there is no way to configure typos here.

PR #1592, opened 2026-08-07
fix(cli): anchored extend-exclude patterns apply to absolute path arguments

### LINKED ISSUE #1075: --extend-exclude doesn't work with absolute paths
I hope this hasn't been reported before.

typos version: `1.23.6`

`/some/path/some_misspellings.txt` is a file with errors which typos in its default configuration would detect.

`/some/path/_typos.toml` has the contents:

```toml
[files]
extend-exclude = [
    '/some_misspellings.txt'
]
```

This works as expected, i.e. no errors are reported:

```bash
cd /some/path/
typos --force-exclude some_misspellings.txt
```

Although this reports spelling errors:

```bash
cd /some/path/
typos --force-exclude /some/path/some_misspellings.txt
```

I assumed that they would lead to the same result.

Eljees:
Fixes #1075.

`extend-exclude` matching is rooted at the literal `"."`:

```rust
let mut ignores = ignore::gitignore::GitignoreBuilder::new(".");
```

`Gitignore` strips the root off each candidate before matching, and a `"."` root can only strip `./`-prefixed candidates. The walker hands back entries in the same spelling as the path argument, so for `typos /abs/dir` (or `--files /abs/dir`) every candidate stays absolute, and anchored patterns like `/skip` never match anything — the exclusion is silently lost. Unanchored patterns keep working because they compile to `**/pat`, which is how this went unnoticed. Reproduced with the 1.49.0 release binaries on both Linux and Windows, identically.

The matcher is now rooted at the invocation directory:

- a relative invocation is byte-for-byte unchanged: relative candidates don't start with the root, the strip is a no-op, exactly as with `"."`;
- an absolute path argument under the CWD is stripped to the same relative form the relative invocation produces, so both spellings of the same directory exclude the same files — which is the contract asked for in the issue thread ("a passed in path should be checked against the root, regardless of format");
- an absolute argument *outside* the CWD is unchanged (nothing to strip, unanchored-only, as today). Making that case work needs an answer to "what is the root" (#593), which this PR deliberately does not touch.

Per CONTRIBUTING, the first commit adds the test pinning the current behaviour (green on the unfixed tree: relative run excludes `skip/`, absolute run of the same directory does not), and the second commit changes the behaviour and flips the assertion. `trycmd` keeps `args` as literals, so an absolute temp path can't go through `tests/cmd/*.toml`; the test uses `assert_fs` + the `CARGO_BIN_EXE_typos` binary, both already available to the suite.

Verified per commit: the new test passes on commit 1 without the fix and on commit 2 with it; `cargo fmt` clean. The full `cli_tests` suite fails 14 of 35 `trycmd` cases in my environment with formatting-only diffs (`-->` vs `╭▸`) — identical on a clean upstream checkout, so it is environmental and unrelated.

Note: this does not touch the `canonicalize` block that #1577 modifies; the change is confined to the matcher root on line 253.

AI-assisted (LLM used for drafting); the runs above are mine.

### DIFF
--- a/crates/typos-cli/src/bin/typos-cli/main.rs
+++ b/crates/typos-cli/src/bin/typos-cli/main.rs
@@ -250,7 +250,13 @@ fn run_checks(args: &args::Args) -> proc_exit::ExitResult {
             walk.sort_by_file_name(|a, b| a.cmp(b));
         }
         if !walk_policy.extend_exclude.is_empty() {
-            let mut ignores = ignore::gitignore::GitignoreBuilder::new(".");
+            // Root the matcher at the invocation directory rather than the literal ".":
+            // candidates coming out of the walker keep the spelling of the path
+            // argument, and `Gitignore` can only strip a "." root off `./`-prefixed
+            // candidates, so for an absolute path argument anchored patterns never
+            // matched (#1075). With the CWD as the root, absolute candidates are
+            // stripped to the same relative form a relative invocation produces.
+            let mut ignores = ignore::gitignore::GitignoreBuilder::new(&global_cwd);
             for pattern in walk_policy.extend_exclude.iter() {
                 ignores
                     .add_line(None, pattern)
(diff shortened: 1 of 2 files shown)


epage:
I'm seeing signs that this is AI generated without sufficient human review and will close before putting more time into this.

PR #1601, opened 2026-08-15
feat(config): support `.config/typos.toml`

### LINKED ISSUE #1445: Support storing config in `.config/typos.toml`
### Description

Hello! I tried looking if this issue was filed before and couldn't find anything. It would be really nice if typos would look for `.config/typos.toml` by default, in addition of the files it's currently looking for.

As someone who likes to keep my workspace clutter-free both physically and digitally, I try as much as possible to move configuration files to a dedicated folder (and advocate for tools to add native support for this 😄). The most common convention is to use `.config`, which builds upon the established convention for `~/.config`.

A "perfect" mirror of the XDG convention would call for `.config/typos/typos.toml`, although I'm unsure this is necessary here (and all tools in the JS ecosystem using cosmiconfig look for files at the root of `.config`). [Mise does support `.config/mise.toml`, `.config/mise/config.toml` and others](https://mise.jdx.dev/configuration.html#mise-toml)

CallumKerson:
Discover configuration inside a project's `.config/` directory, tracking the emerging `.config` directory standard, which specifies `<project>/.config/<toolname>.<ext>`.  See https://github.com/pi0/config-dir.  It is looked up last, after the existing top-level names in the same directory, so existing projects are unaffected and the more discoverable locations will always be preferred. Fixes #1445

### DIFF
--- a/crates/typos-cli/src/config.rs
+++ b/crates/typos-cli/src/config.rs
@@ -14,8 +14,14 @@ pub const SUPPORTED_FILE_NAMES: &[&str] = &[
     PYPROJECT_TOML,
 ];
 
+/// Config file names looked up inside [`CONFIG_DIR`], after [`SUPPORTED_FILE_NAMES`]
+///
+/// See <https://github.com/pi0/config-dir>
+const CONFIG_DIR_FILE_NAMES: &[&str] = &["typos.toml"];
+
 const CARGO_TOML: &str = "Cargo.toml";
 const PYPROJECT_TOML: &str = "pyproject.toml";
+const CONFIG_DIR: &str = ".config";
 
 #[derive(Debug, Clone, Default, serde::Serialize, serde::Deserialize)]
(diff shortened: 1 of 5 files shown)


CallumKerson:
@epage I understand your reservation about hidden config as discussed in #1445, but since `.typos.toml` is already supported supporting `.config/typos.toml` doesn't seem to me to make the problem any worse. My other thought was to implement a CLI flag/env var override for config location, allowing users to put it in `.config/typos.toml` or anywhere else if they want. But that seemed like a much bigger change. If you would like me to try that instead please let me know.

CallumKerson:
also reading your commend here https://github.com/pi0/config-dir/discussions/3 I could add `_config/typos.toml` in addition? With it taking precedence over `.config/`? Please let me know.

epage:
Closing to redirect the converation to #1445 as the conversation is still on whether to support this and not how to implement support, per our contrib guide.

PR #1603, opened 2026-08-16
ci: use the committed lockfile for releases

meop:
## Summary

Add `--locked` to the binary release build so tagged artifacts use the dependency resolution committed with the release.

I left the ordinary test and compatibility jobs unchanged. The existing dedicated `cargo update --workspace --locked` job already enforces the repository's lockfile policy, while the unlocked workspace jobs continue exercising compatible dependency resolution as designed.

## Validation

- `cargo check --locked -p typos-cli --all-targets`
- the existing lockfile CI job remains unchanged

Prepared with assistance from OpenAI Codex; the commit includes a `Co-authored-by` trailer.

### DIFF
--- a/.github/workflows/post-release.yml
+++ b/.github/workflows/post-release.yml
@@ -97,7 +97,7 @@ jobs:
         toolchain: ${{ matrix.rust }}
         targets: ${{ matrix.target }}
     - name: Build release binary
-      run: cargo build --target ${{ matrix.target }} --verbose --release
+      run: cargo build --locked --target ${{ matrix.target }} --verbose --release
     - name: Build archive
       shell: bash
       run: |


epage:
What is this intended to protect against?

meop:
> What is this intended to protect against?

It prevents the build from resolving to newer dependencies than what is in the lockfile.

Since the project already has a lockfile, it should be used to ensure that every build resolves identically; no unreviewed dependency version slips into the release binary, and what is reviewed in the lockfile diff is what actually ships.

epage:
How do you expect us to get in that situation?

meop:
> How do you expect us to get in that situation?

The existing lockfile job protects the normal CI path. Adding --locked to the independently triggered release build also enforces that invariant at the artifact boundary, protecting against a tag bypassing that path or if any of the workflow assumptions in this project change later. This follows the pattern used by many other Rust projects as a defense-in-depth measure.

epage:
Normal CI is gated already. Any "attack" that can bypass that can also bypass the CI job. I'm not convinced why this "defense in depth" will make a difference.

meop:
Sounds good.

Not trying to change your mind, but wanted to clarify that I did not mean "defense-in-depth" in the sense that I thought this project could get attacked.

Typos has this layout:

- cargo update --locked (lockfile job) — .github/workflows/ci.yml
- cargo build --release (release job) — .github/workflows/post-release.yml <-- missing --locked

Many projects do not have an update locked part of their pipeline sequence. Like watchexec project, where its CI did not have build locked either, so it ended up with an unbuildable main branch. Oops.

Most other projects just put build locked in their pipeline, and this forces people to do their own update locked step locally, before committing code. I suppose it could be argued that more projects should instead add update locked as part of their own CI too, like Typos.

So for Typos, yes, today having build locked would be a no-op. But tomorrow, having build locked might protect the build if the update locked step ever got dropped, skipped, or missed in a future CI refactor.

PR #1617, opened 2026-09-10
fix(cli): Don't panic when canonicalize fails mid-walk

### LINKED ISSUE #1444: Panic: walk() should be called first

nightah:
Takes @chatman-media's fix from #1577 and restructures it per your review: the test lands first, passing on `master` via `should_panic`, so the fix shows up as a test diff.

## Cause

`walk_entry` canonicalizes each entry's path before looking up its policy. When that fails the error goes through `report_result`, which reports it and then falls back to `Default::default()`. For a `PathBuf` that is the empty path, which `ConfigEngine` never saw during the walk, so the `policy()` call right after it aborts the process. A release build hits the `expect`, a debug build trips the `debug_assert!` on `path.is_absolute()` just above it, which is why the panic text differs between the two.

## Repro

A directory that is readable but not searchable. `ignore` lists the child, `canonicalize` on it fails with `EACCES`. Deterministic, single threaded, no `sudo`.

```console
$ mkdir -p repro/dir
$ echo 'hello wrold' > repro/dir/a.txt
$ chmod 444 repro/dir
$ cd repro && typos
```

1.50.1:

```console
error: Permission denied (os error 13)
  ─▸ ./dir/a.txt
typos-cli had a problem and crashed. To help us diagnose the problem you can send us a crash report.
[...]
exit: 134
```

With this PR:

```console
error: Permission denied (os error 13)
  ─▸ ./dir/a.txt
exit: 1
```

The error still reports. Only the abort is gone.

#1577 assumed `ignore` catches permission denied first. That holds for an unreadable directory, not a readable non-searchable one: `readdir` succeeds, the entry reaches `walk_entry`, and only `canonicalize` fails.

## Files removed mid-walk

Any `canonicalize` failure lands here, including a file deleted between `readdir` and the lookup. Deleting 4000 files during the walk, 10 runs:

| | 1.50.1 | this PR |
| --- | --- | --- |
| aborted | 10/10 | 0/10 |
| reported `No such file or directory (os error 2)` | 10/10 | 10/10 |

## Fix

`walk_entry` skips the entry after reporting the error instead of carrying on with the empty path, matching how a walk error is already handled a few lines above.

## Commits

`test(cli)` adds `crates/typos-cli/tests/walk_canonicalize.rs`, which deletes the file from inside `filter_entry` so the walker still yields it while `canonicalize` fails. `#[should_panic = ""]` passes on `master`. The expected message is empty because the panic text differs between debug and release.

`fix(cli)` applies the fix and drops the attribute, leaving the assertions that `walk_path` returns `Ok` and that the failure was reported.

Both commits pass `cargo test --workspace`, `cargo fmt --check` and `cargo clippy --workspace --all-targets` standalone.

**EDIT:** History rewritten into one test commit and one fix commit. The Commits section above describes the earlier version, where the test used `should_panic = ""`.

- 25761c7b wraps only the `walk_path` call in `catch_unwind` and asserts the panic message from the failed policy lookup, which differs between debug and release builds, so an unrelated panic like the `unwrap` on the file removal inside the walk no longer satisfies it.
- 160d7928 applies the fix, with the comment on the early return, and switches the test to assert that `walk_path` returns `Ok`.

Fixes #1444  and Closes #1577


crates/typos-cli/src/file.rs
```diff
@@ -960,7 +960,15 @@ fn walk_entry(
             (path, cwd)
         } else {
             let path = entry.path();
-            let abs_path = report_result(path.canonicalize(), Some(path), reporter)?;
+            // Reporting the error and carrying on would leave `abs_path` empty, and
+            // `engine.policy` panics on a path it never saw during the walk.
```
epage: If we have a comment, it likely should go on the `return Ok());` and be simple, like "Avoid a failed `engine.policy` lookup"

crates/typos-cli/tests/walk_canonicalize.rs
```diff
@@ -0,0 +1,76 @@
...
+            self.errors.lock().unwrap().push(format!("{msg:?}"));
+        }
+        Ok(())
+    }
+}
+
+// The panic message depends on the build: a debug build trips the `debug_assert!` on
+// `path.is_absolute()` first, a release build reaches the `expect` below it.
+#[test]
+#[should_panic = ""]
```
epage: This feels like it is trying to meet the letter of the request but not the spirit as someone looking at this can't tell what is panicking. It could be any call or even one of the asserts.

crates/typos-cli/src/file.rs
nightah: Understood will adjust.

nightah: Done in fa59bd92.

crates/typos-cli/tests/walk_canonicalize.rs
nightah: Fair point. c774ae3a reworks it: only the `walk_path` call is wrapped in `catch_unwind`, and the test asserts the panic message from the `engine.policy` lookup. That message depends on the build (debug trips the `debug_assert!` on `path.is_absolute()` first, release reaches the `expect`), so the expected text is picked with `cfg!(debug_assertions)`. It also rules out the `unwrap` on the file removal inside the walk.

To keep the pushed history intact I did this as a revert and re-apply on top rather than rewriting the first two commits. fa59bd92 brings the fix back and switches the test to asserting `Ok`.

nightah:
Let me know if you prefer this addressed in another way.
I just did a revert and re-apply with your proposed changes instead of re-writing history with a force push.

epage:
Our contrib guide asks for atomic commits that reflect how they should be reviewed and merged and now how they were developed. Please rewrite your history.

nightah:
> Our contrib guide asks for atomic commits that reflect how they should be reviewed and merged and now how they were developed. Please rewrite your history.

Adjusted accordingly.

epage:
Thanks!

